PR #1, opened 2021-12-16
Update grpc bindings to use new Vector Service API

benjaminran:
This switches go-pinecone to use the newer external "vector service" api instead of "core.proto", which is no longer intended for external use. The new api is much more user-friendly, as the diff in the sample app hopefully shows.

Planning to tag this as release v0.3.0.

A few more notes about the change:
- added vector_service.proto and removed core.proto/storage.proto, with code generated accordingly
- switched the package name that callers will need to import from pinecone to pinecone_grpc (when we have openapi support too, there would be name conflicts if we put both in the pinecone package; and I don't want to have to move the grpc code out of the pinecone package in future to make place for some more user-friendly wrapper code we may want to add at some point)
- no longer including handpicked protobuf well known types in the google/protobuf subdir, and now using git submodules for the new 3p protobuf dependencies (api-common-protos and grpc-gateway)

Codegen, build, and the sample app all work without any surprises.


examples/app.go
```diff
@@ -14,96 +15,52 @@ import (
...
 	config := &tls.Config{}
 
-	var opts []grpc.DialOption
-	opts = append(opts, grpc.WithTransportCredentials(credentials.NewTLS(config)))
-	opts = append(opts, grpc.WithAuthority(*serverAddr))
-	opts = append(opts, grpc.WithBlock())
-	opts = append(opts, grpc.WithTimeout(5 * time.Second))
-	log.Printf("connecting to %v", *serverAddr)
-	conn, err := grpc.Dial(*serverAddr, opts...)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
jackpertschuk, line 30: nit: spacing `5*time.second()`

jackpertschuk (approved):
nice!

examples/app.go
benjaminran: Will fix and merge :)

PR #11, opened 2024-02-28
Update data plane SDK definition

haruska:
The Pinecone Golang SDK has not been actively developed in over two years. In that time, much has changed in PineconeDB. This PR:

* Updates the data plane protobuf definition to the most current and regenerates the client code.
* Makes the generated code `internal` to the module, reducing the SDK surface area
* Adds the new "pinecone" package that contains the improved SDK definition and implementation
* Adds wrappers around data plane gRPC calls
* Adds test harness and cursory tests for the data plane interface calls
* Adds CI github actions to exec tests on PRs

There is follow-on work that still needs to happen including better testing (see GH Issue #12) and documentation.


pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+
+	return &QueryVectorsResponse{
+		Matches:   matches,
+		Namespace: res.Namespace,
+		Usage:     toUsage(res.Usage),
+	}, nil
+}
+
+func toVector(vector *data.Vector) *Vector {
```
austin-denoble, line 289: **question:** `toVector`, `toScoredVector`, and `vecToGrpc` don't have `nil` checks, which I'm assuming is fine because of how they're used in `for` loops, but would there be any need for that in any of these?

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	namespaceSummaries := make(map[string]*NamespaceSummary)
+	for key, value := range res.Namespaces {
+		namespaceSummaries[key] = &NamespaceSummary{
+			VectorCount: value.VectorCount,
+		}
+	}
```
austin-denoble, line 261: **question:** We do some mapping before returning responses for `FetchVectors`, `ListVectors`, `DescribeIndexStats`, and `query`, and I suppose I'm mainly curious why that is.

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+
+	req := &data.UpsertRequest{
+		Vectors:   vectors,
+		Namespace: in.Namespace,
+	}
+
+	res, err := (*idx.dataClient).Upsert(idx.akCtx(*ctx), req)
+	if err != nil {
+		return 0, err
+	}
```
austin-denoble, line 65: **thought:** I'm not sure what the errors that are returned from the gRPC calls look like, but we may need to eventually think about centralized mapping / wrapping of things if that makes sense in Go.

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+	"crypto/tls"
+	"fmt"
+	"github.com/pinecone-io/go-pinecone/internal/gen/data"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/metadata"
+	"log"
+)
+
+type IndexConnection struct {
```
austin-denoble, line 14: Thanks for pulling context out of here after chatting last week. 👍

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+	idx := IndexConnection{apiKey: apiKey, dataClient: &dataClient, grpcConn: conn}
+	return &idx, nil
+}
+
+func (idx *IndexConnection) Close() error {
+	err := idx.grpcConn.Close()
+	return err
+}
+
+type UpsertVectorsRequest struct {
```
austin-denoble, line 46: **question:** Some structs are defined in-line in this file and some are in `models.go`, would it make sense to keep everything in one place or is it more reasonable to define these things near where they're used?

pinecone/index_connection_test.go
```diff
@@ -0,0 +1,161 @@
+package pinecone
+
+import (
+	"context"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
```
austin-denoble, line 6: Read up a bit on the testing framework in case anyone else is curious: https://github.com/stretchr/testify.

This was the one you mentioned was controversial? Seems fine to me, although I'm not sure if that's something we need to consider at this point.

Also less relevant now, but should we consider having a specific folder path for test files?

austin-denoble (approved):
Nice work here getting this SDK back in business, overall the approach makes sense and feels clean. Caveat to my review being that I'm not that familiar with Go.

Left a lot of questions and nits but I don't necessarily consider any of that blocking. Again, nice job, looking forward to seeing this come together.

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+
+	return &ListVectorsResponse{
+		VectorIds:           vectorIds,
+		Namespace:           res.Namespace,
+		Usage:               &Usage{ReadUnits: res.Usage.ReadUnits},
+		NextPaginationToken: toPaginationToken(res.Pagination),
+	}, nil
+}
+
+type QueryByVectorRequest struct {
```
jhamon, line 142: To be really explicit I would call this `QueryByVectorValuesRequest` and the other `QueryByVectorIdRequest`. In both cases, the query needs a vector to compute similarity.

pinecone/index_connection_test.go
```diff
@@ -0,0 +1,161 @@
...
+	namespace string
+	vectorIds []string
+}
+
+// Runs the test suite with `go test`
+func TestIndexConnection(t *testing.T) {
+	apiKey := os.Getenv("API_KEY")
+
+	podTestSuite := new(IndexConnectionTests)
+	podTestSuite.host = os.Getenv("POD_INDEX_HOST")
```
jhamon, line 26: It looks like these tests expect the index to already exist. Is that handled somewhere in this repo or is that currently a manual setup step?

pinecone/index_connection_test.go
```diff
@@ -0,0 +1,161 @@
...
+	ts.truncateData()
+
+	err := ts.idxConn.Close()
+	assert.NoError(ts.T(), err)
+}
+
+func (ts *IndexConnectionTests) TestFetchVectors() {
+	req := &FetchVectorsRequest{
+		Ids:       ts.vectorIds,
+		Namespace: ts.namespace,
```
jhamon, line 62: My golang is rusty, but is there a way to make Namespace optional in these various request objects? On the one hand I like pushing people to be explicit, but on the other hand historically our other SDKs have treated namespace as optional and automatically applied the default `""` when nothing is passed.

In Typescript we tried pulling the namespace up one level so it's more of a config property baked into the thing analgous to your `IndexConnection`. The advantage of that is the user only has to specify it once and doesn't have to remember to pass the correct namespace on every operation. The thinking was that seeing this param repeated across every method suggests it's really part of the connection config, and having to repeat it on every call feels unnecessarily repetitious and error prone. Food for thought.

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+type DeleteVectorsRequest struct {
+	Namespace string
+	Ids       []string
+	Filter    *Filter
+	DeleteAll bool
+}
+
+func (idx *IndexConnection) DeleteVectors(ctx *context.Context, in *DeleteVectorsRequest) error {
```
jhamon, line 203: I gave similar feedback recently on a java PR, but I think we should consider breaking out some different helper methods for things like `DeleteById`, `DeleteByFilter`, and `DeleteByNamespace` (e.g. DeleteAll) to guide the user better, as the various options accepted by our API are mutually exclusive.

jhamon (approved):
Seems like a good start. Obviously more tests will need to be added over time.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,24 @@
+name: ci
```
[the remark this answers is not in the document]
haruska: unless you feel strongly, I prefer "ci" to "pr" in the tasks are running the CI tests. There could be multiple workflows to execute on PR. Likewise, we could choose to run CI on actions more than just PRs (like commits to main.)

I do agree we should be consistent where possible on SDKs.

justfile
```diff
@@ -0,0 +1,15 @@
...
+  set +o allexport
+  go test ./pinecone
+bootstrap:
+    go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+    go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
+    go install golang.org/x/tools/cmd/godoc@latest
+gen:
+  protoc --experimental_allow_proto3_optional --proto_path=apis/proto --go_opt=module="github.com/pinecone-io/go-pinecone" --go-grpc_opt=module="github.com/pinecone-io/go-pinecone" --go_out=. --go-grpc_out=. apis/proto/pinecone/data/v1/vector_service.proto
+docs:
+  @echo "Serving docs at http://localhost:6060/pkg/github.com/pinecone-io/go-pinecone/pinecone/"
```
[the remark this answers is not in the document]
haruska: Yup, that looks like a good package. I was trying to maintain the functionality that was in the Makefile. Will certainly prep markdown documentation for Mintlify serving. Also need better docstrings everywhere!

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+	}
+	res, err := (*idx.dataClient).List(idx.akCtx(*ctx), req)
+	if err != nil {
+		return nil, err
+	}
+
+	vectorIds := make([]*string, len(res.Vectors))
+	for i := 0; i < len(res.Vectors); i++ {
+		vectorIds[i] = &res.Vectors[i].Id
+	}
```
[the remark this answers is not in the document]
haruska: That's right. This simplifies the result values instead of being wrapped in a struct with one string.

pinecone/index_connection.go
haruska: Good question. I convinced myself that it's not needed but the guards can't hurt. I'll add them.

haruska: https://github.com/pinecone-io/go-pinecone/pull/11/commits/9dbdc9441b8793616ffbcefb8b6a41238da85be0

pinecone/index_connection.go
```diff
@@ -0,0 +1,364 @@
+package pinecone
+
+import (
+	"context"
+	"crypto/tls"
+	"fmt"
+	"github.com/pinecone-io/go-pinecone/internal/gen/data"
```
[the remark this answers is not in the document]
haruska: Yeah, the import/package system in Go works like this. They decided on a "distributed" package manager instead of centralized. The `go.mod` file at root defines this module as `github.com/pinecone-io/go-pinecone`. When it sees that prefix, it's local to this module. Also, `internal` can only be imported by files in the same package.

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+	if sv == nil {
+		return nil
+	}
+	return &data.SparseValues{
+		Indices: sv.Indices,
+		Values:  sv.Values,
+	}
+}
+
+func (idx *IndexConnection) akCtx(ctx context.Context) context.Context {
```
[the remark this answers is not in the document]
haruska: "api key context"  ... as in a new context with the api key added to the metadata.

Other thing of note, anything starting with a lowercase is not exported. So this is a private method on IndexConnection.

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+func (idx *IndexConnection) DeleteVectors(ctx *context.Context, in *DeleteVectorsRequest) error {
+	req := data.DeleteRequest{
+		Ids:       in.Ids,
+		DeleteAll: in.DeleteAll,
+		Namespace: in.Namespace,
+		Filter:    in.Filter,
+	}
+
+	_, err := (*idx.dataClient).Delete(idx.akCtx(*ctx), &req)
+	return err
```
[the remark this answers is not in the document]
haruska: yup. pattern is generally `if err != nil { return nil, err}`. In this case, we only return an `error` (or nil if no error).

This is another cleanup from the generated gRPC calls. The data client returns an empty struct or an error.

pinecone/index_connection.go
haruska: mainly to not expose the `internal` struct types from the generated code. This allows us to have a clean and well documented interface in the `pinecone` package without the user of the SDK having to dig into internal code.

pinecone/index_connection.go
haruska: In Go, every error is an `error`. Unless we wanted different messages, these are generally easier to maintain as pass-through.

pinecone/index_connection.go
haruska: A definite improvement for sure. Thanks for talking through it!

pinecone/index_connection.go
haruska: It's sudo-arbitrary. The exported structs defined here are the Request and Response structs for exported actions. The data structs are in `models.go`. I guess because those have a higher probability of being reused outside of index connection?

This is something that can be re-organized in the future without breaking uses of the SDK. Everything is exported to the `pinecone` package regardless of filename.

pinecone/index_connection_test.go
haruska: Yeah, more that using anything not built in is controversial. This package allows for things like setup and tear-down on suite, sub-suite, test levels. It's just more like other testing frameworks folks are used to. Seemed like a good addition.

justfile
```diff
@@ -0,0 +1,15 @@
+test:
+  #!/usr/bin/env bash
+  set -o allexport
+  source .env
```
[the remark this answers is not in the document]
haruska: 3772901

pinecone/index_connection.go
```diff
@@ -0,0 +1,358 @@
...
+	"log"
+)
+
+type IndexConnection struct {
+	apiKey     string
+	dataClient *data.VectorServiceClient
+	grpcConn   *grpc.ClientConn
+}
+
+func newIndexConnection(apiKey string, host string) (*IndexConnection, error) {
```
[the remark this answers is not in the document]
haruska: Makes sense. This is a non-exported function (index connections are created off of Client). We can add to it as needed without breaking existing clients.

pinecone/index_connection.go
haruska: f354232

pinecone/index_connection_test.go
haruska: The indexes have to already exist with the hosts being in .env. I expect this to change when the control plane api is added to the repo.

pinecone/index_connection_test.go
haruska: So, as-is these are optional. Golang sets any omitted value to the "zero value" of that type. In `string` type's case, that's `""`.

I do like the idea of capturing an optional `Namespace` in the `IndexConnection` struct and removing it from the individual calls. There aren't cases where folks would be changing the namespace per call, right?

jhamon: > There aren't cases where folks would be changing the namespace per call, right?

It's possible they would want to work with multiple namespaces, but based on what I've seen in our docs and examples it's not common. If somebody wanted to work with multiple namespaces they could create two client instances and it wouldn't be a huge deal, certainly less error prone than having to remember to pass a namespace in every method call. Nobody has complained about this choice in the typescript client AFAIK. As an organizational concept, it's kind of a hangover from pre-serverless days when people needed access to a lighter weight form of multitenancy to save the cost of spinning up a completely separate (expensive) pod index.

pinecone/index_connection.go
haruska: 1f9a12d

pinecone/index_connection.go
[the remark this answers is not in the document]
haruska: Will be resolved in a separate PR (GH Issue #13)

pinecone/index_connection_test.go
haruska: a84740c

This cleaned up IndexConnection quite a bit! Note, `IndexConnection.Namespace` is exported and can be changed by the user. The new namespace would then be used in subsequent calls.

PR #14, opened 2024-03-12
Add control plane to SDK

haruska:
This PR adds the Control Plane REST API calls for managing Indexes and Collections. 

Similar to the codegen for the Data Plane gRPC calls, this generates code from the OpenAPI Spec for the Control Plane REST calls. These calls are then wrapped in the SDK to reduce the surface area of the SDK to support.


pinecone/client.go
```diff
@@ -20,3 +41,357 @@ func (c *Client) IndexWithNamespace(host string, namespace string) (*IndexConnec
...
+func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
+	res, err := c.restClient.DeleteIndex(ctx, idxName)
+	if err != nil {
+		return err
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusAccepted {
+		var errResp control.ErrorResponse
+		err = json.NewDecoder(res.Body).Decode(&errResp)
```
ssmith-pc, line 221: Is there a potential resource leak here of the response body not being consumed? 

From https://pkg.go.dev/net/http#:~:text=If%20the%20returned,keep%2Dalive%22%20request.

> If the returned error is nil, the [Response](https://pkg.go.dev/net/http#Response) will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the [Client](https://pkg.go.dev/net/http#Client)'s underlying [RoundTripper](https://pkg.go.dev/net/http#RoundTripper) (typically [Transport](https://pkg.go.dev/net/http#Transport)) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

haruska: The defer line 217 takes care of closing the body. The decoder actually consumes the entire body as well.

pinecone/client.go
```diff
@@ -20,3 +41,357 @@ func (c *Client) IndexWithNamespace(host string, namespace string) (*IndexConnec
...
+		var errorResponse control.ErrorResponse
+		err = json.NewDecoder(res.Body).Decode(&errorResponse)
+		if err != nil {
+			return nil, err
+		}
+
+		return nil, fmt.Errorf("failed to create collection: %s", errorResponse.Error.Message)
+	}
+
+	return decodeCollection(res.Body)
```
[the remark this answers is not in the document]
haruska: Yup. Golang in general is pretty limited with what type of abstractions you can make. One thing to note is anything not capitalized is not exported so these functions do not increase the surface area of the SDK.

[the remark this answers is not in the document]
ssmith-pc: That is good to know

pinecone/client.go
haruska: The status code isn't 202. The body will be a JSON document in the shape of `control.ErrorResponse`. Line 221 passes the `res.Body` to a json decoder and a reference to `errResp` to hold the parsed body. When the body is parsed, it is read.

ssmith-pc: Sorry I wasn't clear -- should have selected the larger code block including return. I am actually asking about the case where it falls through because the Status is not 202 and just returns nil

pinecone/client.go
```diff
@@ -20,3 +41,357 @@ func (c *Client) IndexWithNamespace(host string, namespace string) (*IndexConnec
...
+	metric := control.IndexMetric(in.Metric)
+	req := control.CreateIndexRequest{
+		Name:      in.Name,
+		Dimension: in.Dimension,
+		Metric:    &metric,
+	}
+
+	//add the spec to req.
+	//because this is defined as an anon struct in the generated code, it must match exactly here.
+	req.Spec = control.CreateIndexRequest_Spec{
```
austin-denoble, line 107: I'm sort of not clear on why this is the case as opposed to the `Serverless` variety below. Primarily just not clear on how it's anon in the generated `CreateIndexRequest_Spec`.

pinecone/client.go
```diff
@@ -20,3 +41,357 @@ func (c *Client) IndexWithNamespace(host string, namespace string) (*IndexConnec
...
+			Pods             *int                     `json:"pods,omitempty"`
+			Replicas         *control.PodSpecReplicas `json:"replicas,omitempty"`
+			Shards           *control.PodSpecShards   `json:"shards,omitempty"`
+			SourceCollection *string                  `json:"source_collection,omitempty"`
+		}{
+			Environment:      in.Environment,
+			PodType:          in.PodType,
+			Pods:             in.TotalCount(),
+			Replicas:         in.ReplicaCount(),
+			Shards:           in.ShardCount(),
```
austin-denoble, line 123: My understanding of this and the methods above is we're basically massaging the values passed and setting them to 1 if negative. I had initially thought it was for argument validation.

Are we relying on the server for validations in general?

pinecone/client.go
austin-denoble: I think the `defer res.Body.Close()` deals with it either way, right?

austin-denoble (approved):
Nice work, I think overall this looks good to me. I've left some comments and questions but I don't think there's necessarily anything blocking here. Thanks a lot for getting the testing suite off the ground as part of this work.

I still need to find some time to pull the repo down and get set up to play around locally, but I feel like everything here makes sense.

pinecone/client.go
haruska: Yeah, it's unfortunate on the auto-gen code. On the Serverless autogen, it names the struct. Maybe it's used in more than one place? Either way, I consider it a bug in the generator. Luckily, because of these wrappers, folks using the SDK don't have to deal with this. 

As far as I know this is the only way in Go. I tried naming the `struct` shape locally but it considered it two different types.

pinecone/client.go
haruska: Another unfortunate case. These are optional params. If you don't set them in the `in` struct, they are the "zero value" for that type. In most cases, the "zero value" for the type works out OK (empty string, 0, false, etc). However, the API defaults these to 1. So, if the caller of the SDK omits it (or sets it to the zero value) we need to ensure the default is passed in.

The other annoyance that is cleaned up here is one of general "Pod" math by our API.  Pods must equal "Replicas * Shards" where Replicas and Shards are both ints with a min value of 1. This only accepts Replicas and Shards as inputs, omitting the ability to miscalculate Pods.

pinecone/client.go
```diff
@@ -1,12 +1,33 @@
...
 type Client struct {
-	apiKey string
+	apiKey     string
+	restClient *control.Client
 }
 
-func NewClient(apiKey string) *Client {
-	c := Client{apiKey: apiKey}
-	return &c
+func NewClient(apiKey string) (*Client, error) {
```
jhamon, line 18: We will definitely want to expand the options supported by the client in the future (for example, for proxy and SSL configurations similar to what I've been implementing in python this week). So some thought about if there's a way to handle future optional configs without creating a breaking change might be worthwhile.

jhamon (approved):

apis/openapi/control/v1/control_v1.yaml
```diff
@@ -0,0 +1,1136 @@
+openapi: 3.0.3
```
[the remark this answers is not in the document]
haruska: I guess inadvertantly. The codegen requires it. In the Protos case for Golang the generated code actually references the proto file it was generated from. Your IDE then uses that as reference.

pinecone/client.go
austin-denoble: Thanks for the explanation, I actually really like the enforcement of "Pod" math through this.

apis/openapi/control/v1/control_v1.yaml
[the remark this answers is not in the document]
haruska: removed to a private submodule.

pinecone/client.go
```diff
@@ -1,12 +1,33 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"github.com/deepmap/oapi-codegen/v2/pkg/securityprovider"
+	"github.com/pinecone-io/go-pinecone/internal/gen/control"
+	"io"
+	"net/http"
+)
+
 type Client struct {
```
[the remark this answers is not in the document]
haruska: No, you can import "github.com/pinecone-io/go-pinecone/pinecone" So:

```go
package main

import (
	"context"
	"fmt"
	"github.com/pinecone-io/go-pinecone/pinecone"
)

func main() {
	pc := pinecone.NewClient("api-key")
	idx, err := pc.Index("index-host")
	defer idx.Close()
	
	if err != nil {
		fmt.Println("Error:", err)
		return
	}
	
	ctx := context.Background()
	
	res, err := idx.DescribeIndexStats(&ctx)
	if err != nil {
		fmt.Println("Error:", err)
		return
	}
	
	fmt.Println(res)
}
```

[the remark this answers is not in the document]
haruska: I believe `Client` must be exported. The fields `apiKey` and `restClient` however are not exported. The only real way to get a valid client is through `NewClient`

pinecone/client.go
```diff
@@ -1,12 +1,33 @@
...
-func NewClient(apiKey string) *Client {
-	c := Client{apiKey: apiKey}
-	return &c
+func NewClient(apiKey string) (*Client, error) {
+	apiKeyProvider, err := securityprovider.NewSecurityProviderApiKey("header", "Api-Key", apiKey)
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := control.NewClient("https://api.pinecone.io", control.WithRequestEditorFn(apiKeyProvider.Intercept))
```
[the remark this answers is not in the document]
haruska: It does not make a network call. I would think it highly unlikely an `error` would be returned here. 

The generated code signature for `control.NewClient` does allow for it and should be handled. Apparently, providing `ClientOptions` to `NewClient` is able to return an `error`.

README.md
```diff
@@ -33,6 +33,11 @@ To get help using go-pinecone, reach out to support@pinecone.io.
 
 Then, execute `just bootstrap` to install the necessary Go packages
 
+### .env Setup
+
+To avoid race conditions or having to wait for index creation, the tests require a project with at least one pod index
+and one serverless index. Copy the api key and index names to a `.env` file. See `.env.example` for a template.
+
```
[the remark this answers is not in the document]
haruska: Updated env vars at 4bf6238

pinecone/client.go
haruska: Good callout. I've changed to an input struct which will allow for additional params in the future without breaking current usages.

4d4dc0c

PR #15, opened 2024-03-19
Management Plane Alpha: Project and API Key Management

haruska:
Adds the management plane alpha API to the Go SDK. Implementation is similar to the control plane client currently in stable.

This initial alpha release provides robust support for managing projects and their associated API keys directly through the Go client library.

New Features

Project Management

* CreateProject: Allows users to create new projects, specifying project names.
* ListProjects: Enables listing all projects available to the user.
* FetchProject: Retrieves details for a specific project by its ID.
* DeleteProject: Removes a project identified by its ID.

API Key Management

* CreateApiKey: Supports creating new API keys within a specified project, including setting names for easy identification.
* ListApiKeys: Lists all API keys associated with a given project, enhancing manageability and oversight.
* FetchApiKey: Simplified to fetch details of an API key using its ID only, removing the need for projectId in the request.
* DeleteApiKey: Allows for the deletion of specific API keys by their ID, completing the set of CRUD operations for API key management.

---
  - https://app.asana.com/0/0/1206859452040361


austin-denoble (approved):
Nice work. straightforward and easy to understand, I've got some questions and nits but nothing blocking. Thanks a bunch for adding in docstring comments for the management plane stuff as a part of implementation. 🥳

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+//
+//	clientParams := NewManagementClientParams{ApiKey: "your_api_key_here"}
+//	managementClient, err := NewManagementClient(clientParams)
+//	if err != nil {
+//	    log.Fatalf("Failed to create management client: %v", err)
+//	}
+//	// Now you can use managementClient to interact with the API
+//
+// Note that ManagementClient methods are designed to be safe for concurrent use by multiple
+// goroutines, assuming that its configuration (e.g., the API key) is not modified after
```
aulorbe, line 44: Is there a way to make the API key protected to ensure this? Something like Java's `private`?

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+//
+// Example:
+//
+//	projects, err := managementClient.ListProjects(ctx)
+//	if err != nil {
+//	    log.Fatalf("Failed to list projects: %v", err)
+//	}
+//	for _, project := range projects {
+//	    fmt.Printf("Project ID: %s, Name: %s\n", project.Id, project.Name)
+//	}
```
aulorbe, line 125: srsly fantastic comments

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+	if res.JSON500 != nil {
+		return nil, fmt.Errorf("internal server error: %s", res.JSON500.Error.Message)
+	}
+
+	if res.JSON4XX != nil {
+		return nil, fmt.Errorf("unknown error: %s", res.JSON4XX.Error.Message)
+	}
+
+	if res.JSON200 == nil || res.JSON200.Data == nil {
+		return nil, fmt.Errorf("unexpected response format or empty data")
```
aulorbe, line 145: This might not be a thing that's idiomatic in Go, but would it be easier to extract these mappings into a separate function? Then you can make this function basically like 

```
If error not 2xx:
    callTheErrorMappingFunction(errorMessage)
```

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+//
+// The ManagementClient is designed to simplify the management of projects and
+// their API keys by abstracting away the direct handling of HTTP requests and
+// responses. It leverages a generated low-level client (`restClient`) for
+// communication with the API, ensuring that requests are properly authenticated
+// and formatted according to the API's specification.
+//
+// Fields:
+//   - apiKey: The API key used for authenticating requests to the management API.
+//     This key should have the necessary permissions for the operations you intend to perform.
```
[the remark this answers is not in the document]
haruska: it's tough because org keys aren't generally available yet... so there's nothing to point to publicly. I'm open to suggestions. I think this general confusion exists when discussing the Management API in general.

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+//
+// Example:
+//
+//	project, err := managementClient.FetchProject(ctx, "your_project_id_here")
+//	if err != nil {
+//	    log.Fatalf("Failed to fetch project: %v", err)
+//	}
+//	fmt.Printf("Project ID: %s, Name: %s\n", project.Id, project.Name)
+func (c *ManagementClient) FetchProject(ctx context.Context, projectId uuid.UUID) (*Project, error) {
+	resp, err := c.restClient.FetchProjectWithResponse(ctx, projectId)
```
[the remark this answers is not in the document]
aulorbe: ⬆️

aulorbe (approved):
Just some Qs for my understanding and learning :)

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+// initialization.
+type ManagementClient struct {
+	apiKey     string
+	restClient *management.ClientWithResponses
+}
+
+// NewManagementClientParams holds the parameters for creating a new ManagementClient.
+type NewManagementClientParams struct {
+	ApiKey string
+}
```
ssmith-pc, line 54: I'm guessing we're going to want SourceTag here as well, and the Interceptor to set it on the client

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+//
+// It is important to handle the error returned by this method to ensure that the
+// management client has been created successfully before attempting to make API calls.
+func NewManagementClient(in NewManagementClientParams) (*ManagementClient, error) {
+	apiKeyProvider, err := securityprovider.NewSecurityProviderApiKey("header", "Api-Key", in.ApiKey)
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := management.NewClientWithResponses("https://api.pinecone.io/management/v1alpha", management.WithRequestEditorFn(apiKeyProvider.Intercept))
```
ssmith-pc, line 90: For our integration tests, we may want to be able to specify an alternate API endpoint. Consider adding an environment variable to allow that?

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+// ListProjects retrieves all projects from the management API. It makes a call to the
+// management plane's ListProjects endpoint and returns a slice of Project pointers.
+//
+// The method handles various HTTP response codes from the API, including:
+// - 401 Unauthorized: Returned if the API key is invalid or missing.
+// - 500 Internal Server Error: Indicates a server-side error. It might be temporary.
+// - 4XX: Covers other client-side errors not explicitly handled by other conditions.
+//
+// In case of a successful response but with unexpected format or empty data,
+// it returns an error indicating the unexpected response format.
```
ssmith-pc, line 108: To me this seems like something that is important to check for within the function impl, but is distracting/confusing to document as behavior in the client. 

The user will not care about the distinction that the server told us the request succeeded but gave us no data back. It's just a server error from their perspective.

Maybe just move this comment down to where that condition is being checked.

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+		}
+	case http.StatusUnauthorized:
+		return nil, fmt.Errorf("unauthorized: %v", resp.JSON401)
+	case http.StatusNotFound:
+		return nil, fmt.Errorf("project not found: %v", resp.JSON404)
+	case http.StatusInternalServerError:
+		return nil, fmt.Errorf("internal server error: %v", resp.JSON500)
+	default:
+		return nil, fmt.Errorf("unexpected status code: %d", resp.StatusCode())
+	}
```
ssmith-pc, line 199: In the default case, is there any additional error message we can include from the response body? Otherwise the client will only see `"unexpected status code: 4xx"`, but any other details will be swallowed.

pinecone/management_client.go
```diff
@@ -0,0 +1,488 @@
...
+//	}
+func (c *ManagementClient) DeleteProject(ctx context.Context, projectId uuid.UUID) error {
+	resp, err := c.restClient.DeleteProjectWithResponse(ctx, projectId)
+	if err != nil {
+		return fmt.Errorf("failed to delete project: %w", err)
+	}
+
+	switch resp.StatusCode() {
+	case http.StatusOK, http.StatusAccepted, http.StatusNoContent:
+		return nil // Success case
```
ssmith-pc, line 275: is this because we don't know what the server success code is going to be, or we just don't care as long as it's `2xx`?

pinecone/management_client_test.go
```diff
@@ -0,0 +1,146 @@
...
+	require.Greater(ts.T(), len(apiKeys), 0, "API key list in test setup should not be empty")
+	ts.apiKey = *apiKeys[0]
+}
+
+func (ts *ManagementClientTests) TestListProjects() {
+	projects, err := ts.client.ListProjects(context.Background())
+	require.NoError(ts.T(), err, "Failed to list projects")
+	require.Greater(ts.T(), len(projects), 0, "Projects list should not be empty")
+}
+
```
ssmith-pc, line 53: Are all of these tests happy path? Should we add some sad path test cases?

ssmith-pc (approved):
I will echo others' feedback that it's great how well written this is. Added a few comments, don't think it's anything blocking

pinecone/management_client.go
haruska: Unfortunately, the generated "response" types are all different for each call. Better to just inline it

pinecone/management_client.go
haruska: Good call. Added at adbc16e 

@ssmith-pc mind giving a quick look over?

pinecone/management_client.go
haruska: 675ff33

pinecone/management_client.go
haruska: good callout. Updated default on all calls to return the response body as well: 9692885

pinecone/management_client.go
haruska: More we don't care. Also, it's based on an alpha stability API. any `2xx` works for us.

pinecone/management_client.go
```diff
@@ -0,0 +1,500 @@
...
+// Returns a pointer to an initialized ManagementClient instance on success. In case of
+// failure, it returns nil and an error describing the issue encountered. Possible errors
+// include issues with setting up the API key provider or problems initializing the
+// underlying REST client.
+//
+// Example:
+//
+//	clientParams := NewManagementClientParams{
+//	    ApiKey: "your_api_key_here",
+//	    SourceTag: "my-application", // Optional but recommended for identifying request sources
```
ssmith-pc, line 92: And here I'd just leave this commented as `// Optional`

pinecone/management_client.go
```diff
@@ -0,0 +1,500 @@
...
+// This method sets up the management plane client with the necessary configuration for
+// authentication and communication with the management API.
+//
+// The method requires an input parameter of type NewManagementClientParams, which includes:
+//   - ApiKey: The API key used for authenticating requests to the management API.
+//     This key should have the necessary organization-level permissions for the operations
+//     you intend to perform.
+//   - SourceTag: An optional string used to identify the source of the requests. This tag
+//     is included in the User-Agent header of each request made by this client, providing
+//     a way to trace and analyze the usage patterns or origins of API requests.
```
ssmith-pc, line 78: For reference, here is the documentation we'll point our partners to about setting up attribution for their integration: https://docs.pinecone.io/integrations/build-integration/attribute-api-activity

Source Tag is a tricky one to document because it's only conditionally recommended: if you are part of our integration partner program then it's recommended, but we also don't want to make anyone feel bad, like they're not cool enough to be a partner so don't bother setting that field.

I think here we'd probably just want to limit what we say about it and instead reference the Partner documentation. Something like:

```
SourceTag: An optional string used to help Pinecone attribute API activity to our partners. 
For more info, see https://docs.pinecone.io/integrations/build-integration/attribute-api-activity"
```

pinecone/management_client.go
ssmith-pc: Setting the header looks good, added some comments on the comments

pinecone/management_client.go
haruska: Updated at 72c9f3a.

pinecone/management_client.go
haruska: Updated at 72c9f3a.

pinecone/management_client_test.go
haruska: Yeah, this is a great callout. The entire test suite could use more robust testing. It's difficult to trigger some of the sad paths since these are hitting live servers. Should be holistically considered before v1.0

pinecone/management_client.go
haruska: f691c36

PR #16, opened 2024-03-26
Adds optional source_tag to User-Agent

ssmith-pc:
Allows setting source_tag in UserAgent.

Usage:
```go
	client, err := pinecone.NewClient(pinecone.NewClientParams{
		ApiKey: "my-api-key",
		SourceTag: "foo",
	})
```

### DIFF
--- a/internal/util/user_agent.go
+++ b/internal/util/user_agent.go
@@ -1,0 +1,44 @@
+package util
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+func BuildUserAgent(sourceTag string) string {
+	return buildUserAgent("go-client", sourceTag)
+}
+
+func BuildUserAgentGRPC(sourceTag string) string {
+	return buildUserAgent("go-client[grpc]", sourceTag)
+}
+
+func buildUserAgent(appName string, sourceTag string) string {
+	// need to set to actual current version
+	appVersion := "0.0.1"
+
+	sourceTagInfo := ""
+	if sourceTag != "" {
+		sourceTagInfo = buildSourceTagField(sourceTag)
+	}
+	userAgent := fmt.Sprintf("%s/%s%s", appName, appVersion, sourceTagInfo)
+	return userAgent
+}
+
+func buildSourceTagField(userAgent string) string {
+	// Lowercase
+	userAgent = strings.ToLower(userAgent)
+
+	// Limit charset to [a-z0-9_ ]
+	re := regexp.MustCompile(`[^a-z0-9_ ]`)
+	userAgent = re.ReplaceAllString(userAgent, "")
+
+	// Trim left/right whitespace
+	userAgent = strings.TrimSpace(userAgent)
+
+	// Condense multiple spaces to one, and replace with underscore
+	userAgent = strings.Join(strings.Fields(userAgent), "_")
+
+	return fmt.Sprintf("; source_tag=%s;", userAgent)
+}
--- a/pinecone/client.go
+++ b/pinecone/client.go
@@ -6,7 +6,8 @@ import (
 	"fmt"
 	"github.com/deepmap/oapi-codegen/v2/pkg/securityprovider"
 	"github.com/pinecone-io/go-pinecone/internal/gen/control"
-	"github.com/pinecone-io/go-pinecone/internal/util"
+	"github.com/pinecone-io/go-pinecone/internal/provider"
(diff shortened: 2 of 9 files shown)


pinecone/client.go
```diff
@@ -6,17 +6,34 @@ import (
...
+	userAgent string
+}
+
+func NewUserAgentProvider(userAgent string) *CustomHeader {
+	return &CustomHeader{userAgent: userAgent}
+}
+
+func (s *CustomHeader) Intercept(ctx context.Context, req *http.Request) error {
+	req.Header.Set("User-Agent", s.userAgent)
+	return nil
```
ssmith-pc, line 30: This approach is based on https://medium.com/@kyodo-tech/go-client-code-generation-from-swagger-and-openapi-a0576831836c

pinecone/client.go
```diff
@@ -6,17 +6,34 @@ import (
...
 }
 
 type NewClientParams struct {
 	ApiKey string
+	// optional fields
+	SourceTag  string
```
ssmith-pc, line 36: Not sure how to include an "optional" field

pinecone/client.go
```diff
@@ -6,17 +6,34 @@ import (
...
 	"github.com/pinecone-io/go-pinecone/internal/gen/control"
+	"github.com/pinecone-io/go-pinecone/internal/util"
 	"io"
 	"net/http"
 )
 
 type Client struct {
 	apiKey     string
 	restClient *control.Client
+	sourceTag  string
```
ssmith-pc, line 17: This field is optional, not sure how to indicate that

pinecone/index_connection.go
```diff
@@ -18,14 +19,15 @@ type IndexConnection struct {
...
-func newIndexConnection(apiKey string, host string, namespace string) (*IndexConnection, error) {
+func newIndexConnection(apiKey string, host string, namespace string, sourceTag string) (*IndexConnection, error) {
 	config := &tls.Config{}
 	target := fmt.Sprintf("%s:443", host)
 	conn, err := grpc.Dial(
 		target,
 		grpc.WithTransportCredentials(credentials.NewTLS(config)),
 		grpc.WithAuthority(target),
 		grpc.WithBlock(),
+		grpc.WithUserAgent(util.BuildUserAgentGRPC(sourceTag)),
```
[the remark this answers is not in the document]
haruska: looks like a failing  test on this change.

haruska (commented):
Approach seems ok. I'd adjust the surface area exposed in the SDK a bit.

pinecone/client.go
ssmith-pc: Good call, I moved it to an internal `provider` package. I also changed it to be generic to setting any header

pinecone/client.go
ssmith-pc: good to know, thanks

internal/util/user_agent.go
```diff
@@ -0,0 +1,44 @@
+package util
```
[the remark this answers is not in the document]
ssmith-pc: renamed -- have a look at the new package and file names. I hate naming things

pinecone/index_connection.go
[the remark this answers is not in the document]
ssmith-pc: fixed (still planning to add more tests, but fixed this existing one)

pinecone/client.go
```diff
@@ -6,17 +6,22 @@ import (
...
 }
 
 type NewClientParams struct {
 	ApiKey string
+	// optional fields
+	SourceTag string
```
haruska, line 24:
```suggestion
	SourceTag string // optional
```

haruska (changes requested):
Looks good. Nice unit tests on setting the custom headers in `internal`. 

I'd add tests for creating a `Client` with and without a `SourceTag`. Same with creating an `IndexConnection` off those clients.

ssmith-pc:
> I'd add tests for creating a `Client` with and without a `SourceTag`. Same with creating an `IndexConnection` off those clients.

Will do. I was wondering if there is a way to intercept the requests and validate the source tag is set

pinecone/client.go
tdonia: unless there's a perf reason (which i don't think there is) we probably want to avoid pointers here as they tend to confuse newbies

tdonia (approved):
looks good!

haruska (approved):
:+1:

PR #18, opened 2024-04-30
Optional configs for setting custom headers / metadata for REST / gRPC operations

austin-denoble:
## Problem
Currently, the Go SDK does not support specifying custom headers or metadata for control & data plane operations (REST and gRPC). This is a useful feature to aid in debugging or tracking a specific request, and we'd like to enable this functionality in the Go SDK.

## Solution
- Update `NewClientParams` and `Client` structs to support `headers`, also allow passing a custom `RestClient` to the new client. This is primarily for allowing mocking in unit tests, but we have a similar approach in other clients allowing users to customize the HTTP module.
- Add new `buildClientOptions` function.
- Update `NewClient` to support appending `Headers` and `RestClient` to `clientOptions`.
- Add error message for empty `ApiKey` without any `Authorization` header provided. If the user passes both, avoid applying the `Api-Key` header in favor of the `Authorization` header. 
- Add new `Client.IndexWithAdditionalMetadata` constructor.
- Update `IndexConnection` struct to include `additionalMetadata`, and update `newIndexConnection` to accept `additionalMetadata` as an argument.
- Update `IndexConnection.akCtx` to handle appending the `api-key` and any `additionalMetadata` to requests.
- Update unit tests, add new `mocks` package and `mock_transport` to facilitate validating requests made via `Client` including.

I spent some time trying to figure out how to mock / test `IndexConnection` but I think it'll be a bit trickier than `Client`, so somewhat saving for a future PR atm.

## Type of Change

## Test Plan
`just test` to run the test suite locally. Validate tests pass in CI.


pinecone/client.go
```diff
@@ -4,51 +4,68 @@ import (
...
 
 func NewClient(in NewClientParams) (*Client, error) {
+	clientOptions := []control.ClientOption{}
 	apiKeyProvider, err := securityprovider.NewSecurityProviderApiKey("header", "Api-Key", in.ApiKey)
 	if err != nil {
 		return nil, err
 	}
-
+	clientOptions = append(clientOptions, control.WithRequestEditorFn(apiKeyProvider.Intercept))
+
```
haruska, line 37: Is an API Key now optional with the addition of custom headers? Still seems required here.

If a caller was attempting to use a JWT Bearer token instead of an API Key for auth, would this now work? Seems I'd need to pass in an API Key (which gets added to the headers) *and* a custom header with the Bearer token. Also guessing that breaks the request (two auth schemes in the same request headers)

haruska (commented):
Good addition overall. Might want to consider some breaking changes pre-1.0 to clean up some of the options in the constructors.

pinecone/client_test.go
```diff
@@ -66,8 +68,11 @@ func (ts *ClientTests) TestNewClientParamsSet() {
 	if client.sourceTag != "" {
 		ts.FailNow(fmt.Sprintf("Expected client to have empty sourceTag, but got '%s'", client.sourceTag))
 	}
+	if client.headers != nil {
+		ts.FailNow(fmt.Sprintf("Expected client to have nil headers, but got '%v'", client.headers))
```
aulorbe, line 72: Total nit, but when I read this, I ask myself "what are nil headers" because it sounds like a quantity. Maybe rephrase this stmt to be `"Expected client headers to be nil"`.

aulorbe (approved):
Just lurking :) Giving an approval so I don't block!

pinecone/client.go
```diff
@@ -65,7 +82,7 @@ func (c *Client) ListIndexes(ctx context.Context) ([]*Index, error) {
 	if res.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("unexpected status code: %d", res.StatusCode)
 	}
-
+	fmt.Printf("res.Body: %+v", res.Body)
```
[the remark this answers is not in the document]
austin-denoble: Ah yes, thank you.

pinecone/client.go
```diff
@@ -4,51 +4,68 @@ import (
...
+	return c.IndexWithAdditionalMetadata(host, "", nil)
 }
 
 func (c *Client) IndexWithNamespace(host string, namespace string) (*IndexConnection, error) {
-	idx, err := newIndexConnection(c.apiKey, host, namespace, c.sourceTag)
+	return c.IndexWithAdditionalMetadata(host, namespace, nil)
+}
+
+func (c *Client) IndexWithAdditionalMetadata(host string, namespace string, additionalMetadata map[string]string) (*IndexConnection, error) {
+	idx, err := newIndexConnection(c.apiKey, host, namespace, c.sourceTag, additionalMetadata)
```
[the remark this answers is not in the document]
austin-denoble: Agreed, will go ahead and take a pass on that now.

pinecone/client.go
austin-denoble: Good point, I didn't change anything regarding API key being optional vs. required so I think it still expects it to be populated.

The point about JWT Bearer is intersting, I'd maybe need to defer to @jhamon regarding specifically what we need to support here, but maybe I should look at removing the API key if they've passed a header for auth, for example. 🤔

justfile
```diff
@@ -3,7 +3,7 @@ test:
   set -o allexport
   source .env
   set +o allexport
-  go test -count=1 ./pinecone
+  go test -count=1 -v ./pinecone
```
austin-denoble, line 6: I like the additional logging output when running these locally, etc.

pinecone/client.go
```diff
@@ -408,3 +412,40 @@ func minOne(x int32) int32 {
...
+
+	for key, value := range in.Headers {
+		headerProvider := provider.NewHeaderProvider(key, value)
+
+		if strings.Contains(key, "Authorization") {
+			hasAuthorizationHeader = true
+		}
+
+		clientOptions = append(clientOptions, control.WithRequestEditorFn(headerProvider.Intercept))
+	}
```
ssmith-pc, line 432: Are there any headers we'd want to prevent them from setting? I'm not sure there, just asking the question.

pinecone/index_connection.go
```diff
@@ -16,19 +16,28 @@ import (
...
-func newIndexConnection(apiKey string, host string, namespace string, sourceTag string) (*IndexConnection, error) {
+type newIndexParameters struct {
+	apiKey string
+	host string
+	namespace string
+	sourceTag string
+	additionalMetadata map[string]string
+}
+
+func newIndexConnection(in newIndexParameters) (*IndexConnection, error) {
```
ssmith-pc, line 32: Nice, I like this

pinecone/models.go
```diff
@@ -112,4 +112,4 @@ type Usage struct {
 }
 
 type Filter = structpb.Struct
-type Metadata = structpb.Struct
+type Metadata = structpb.Struct
```
ssmith-pc, line 115: accidental change?

pinecone/client.go
```diff
@@ -4,51 +4,55 @@ import (
...
 	apiKey     string
 	restClient *control.Client
 	sourceTag  string
+	headers    map[string]string
 }
 
 type NewClientParams struct {
-	ApiKey    string
-	SourceTag string // optional
+	ApiKey     string            // optional unless no Authorization header provided
```
ssmith-pc, line 25: Suggest rephrasing this to "required unless Authorization header provided"

pinecone/client.go
```diff
@@ -408,3 +412,40 @@ func minOne(x int32) int32 {
...
+
+	userAgentProvider := provider.NewHeaderProvider("User-Agent", useragent.BuildUserAgent(in.SourceTag))
+	clientOptions = append(clientOptions, control.WithRequestEditorFn(userAgentProvider.Intercept))
+
+	for key, value := range in.Headers {
+		headerProvider := provider.NewHeaderProvider(key, value)
+
+		if strings.Contains(key, "Authorization") {
+			hasAuthorizationHeader = true
+		}
```
ssmith-pc, line 429: http headers are typically case in-sensitive so we probably want to relax this check

pinecone/client.go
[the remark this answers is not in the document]
austin-denoble: See `newIndexParameters` struct type which `newIndexConnection` should consume.

pinecone/index_connection.go
austin-denoble: Props to @haruska for the suggestion.

pinecone/models.go
austin-denoble: I think it was just auto-formatting removing the line at the end of the file. Is that an issue here?

pinecone/client.go
austin-denoble: At the moment I don't think there are, good question though I'm also not sure.

pinecone/client.go
austin-denoble: Updated to use `strings.ToLower(key), "authorization")`, thanks!

ssmith-pc (approved):

Issue #19, opened 2024-04-30
[Bug] Unusual use of *context.Context


robfig:
### Is this a new bug?

- [X] I believe this is a new bug
- [X] I have searched the existing issues, and I could not find an existing issue for this bug

### Current Behavior

Updating from v0.3 to v0.4, I like the new API, much easier to use. Nice work!

However, I did notice that the API methods all accept a `*context.Context` instead of a `context.Context`. That is highly unusual. Every other usage of context.Context inside the stdlib and other API clients is `context.Context`, which makes me think it was unintentional. Taking address-of to satisfy the interface is no big deal, but it does seem like an API design error. 

### Expected Behavior

Use context.Context, instead of a pointer to that interface.

### Steps To Reproduce

N/A

### Environment

```markdown
- **OS**:
- **Language version**:
- **Pinecone client version**:
```

austin-denoble:
Thanks for the callout, @robfig. I think you're right, I've got a PR up to tweak the API: https://github.com/pinecone-io/go-pinecone/pull/20

Props to @haruska for getting the improved v0.4 client released!

austin-denoble:
`v0.5.0` has been released which should address this. Thanks again! https://github.com/pinecone-io/go-pinecone/releases/tag/v0.5.0

PR #20, opened 2024-05-01
Receive and pass `context.Context` by value in `IndexConnection` methods

austin-denoble:
## Problem
From this issue: https://github.com/pinecone-io/go-pinecone/issues/19

We're receiving and passing `*context.Context` in `IndexConnection` methods. This differs from our `Client` methods and is non-standard for how most libraries handle `Context`.

## Solution
- Swap to receiving and passing `context.Context` by value rather than a pointer in `IndexConnection` methods, update README example to account.

## Type of Change

## Test Plan
CI Testing, make sure you can `just test` locally or `go build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 > **Warning**
-> 
+>
 > **Under active development** This SDK is pre-1.0 and should be considered unstable. Before a 1.0 release, there are
 > no guarantees of backward compatibility between minor versions.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
 > **Warning**
->
```
austin-denoble, line 2: Prettier did a bunch of auto-formatting on some of this locally, sorry about that. 😬

pinecone/index_connection.go
```diff
@@ -46,7 +47,7 @@ func (idx *IndexConnection) Close() error {
 	return err
 }
 
-func (idx *IndexConnection) UpsertVectors(ctx *context.Context, in []*Vector) (uint32, error) {
+func (idx *IndexConnection) UpsertVectors(ctx context.Context, in []*Vector) (uint32, error) {
```
austin-denoble, line 50: I think this is fine, especially given `akCtx` returns by value. At least that's my understanding.

haruska: Double check and make sure that it isn't doing a copy on the stack. The idea is you should be able to pass a Context to different function calls and do a "Cancel" on the single context to abort all the in-process calls.

If it's copying the Context instead of passing a reference that wouldn't work.

haruska (approved):
Looks good! Just double check on the `Context` copy.

PR #21, opened 2024-05-03
Add `json:""` marshaling annotations to response structs

austin-denoble:
## Problem
There's been some UX feedback regarding marshaling of some of our response structs.

Primarily, we're not using `json:"key_name"` annotation for our struct fields so they default to capitalized keys in JSON output when marshaled. IE: `{ "Usage": { "ReadUnits": 5 }}`. We'd like these lower case and matched with our API spec.

There are also instances where we're returning pointers for certain `*int32` fields which can be irritating to consumers needing to dereference and nil-check. In certain cases we can default any nil pointer values to 0 before returning things.

There's no testing for the marshaling on any of our custom structs.

## Solution
I may have gone a bit overboard with the testing here. The bulk of this is tests and the tests are all very similar, definitely open to opinions on removing if people find them unnecessary, but changing up how the JSON is produced is pretty easy and it's an important point of integration in my mind, so maybe worth testing for.

I also wasn't clear if adding these annotations to our request structs made sense either, so I left those out for now. Overall I'm still working on getting a better understanding of Go syntax and expectations.

- Add JSON marshaling annotations to the response structs in `models.go` and `index_connection.go`.
- Update `Collection` and `Usage` structs to default integer values rather than using pointers.
- Add new `models_test.go` file to run through some validation on struct marshaling, also add unit tests to `index_connection_test.go`.

Originally I thought `omitempty` for some of these pointers may be helpful from a consumer standpoint, but I'm actually not sure the more I've thought about it. Does it make sense to omit more or less things? Other feedback here would be much appreciated.

## Type of Change

## Test Plan
Unit tests to explicitly stress marshaling the structs via `json.Marshal()`.

### DIFF
--- a/pinecone/index_connection.go
+++ b/pinecone/index_connection.go
@@ -14,18 +14,18 @@ import (
 )
 
 type IndexConnection struct {
-	Namespace  string
-	apiKey     string
+	Namespace          string
+	apiKey             string
 	additionalMetadata map[string]string
-	dataClient *data.VectorServiceClient
-	grpcConn   *grpc.ClientConn
+	dataClient         *data.VectorServiceClient
+	grpcConn           *grpc.ClientConn
 }
 
 type newIndexParameters struct {
-	apiKey string
-	host string
-	namespace string
-	sourceTag string
+	apiKey             string
+	host               string
+	namespace          string
+	sourceTag          string
 	additionalMetadata map[string]string
 }
 
@@ -94,6 +94,7 @@ func (idx *IndexConnection) FetchVectors(ctx context.Context, ids []string) (*Fe
 	for id, vector := range res.Vectors {
 		vectors[id] = toVector(vector)
 	}
+	fmt.Printf("VECTORS: %+v\n", vectors)
 
 	return &FetchVectorsResponse{
 		Vectors: vectors,
@@ -132,7 +133,7 @@ func (idx *IndexConnection) ListVectors(ctx context.Context, in *ListVectorsRequ
 
 	return &ListVectorsResponse{
 		VectorIds:           vectorIds,
-		Usage:               &Usage{ReadUnits: res.Usage.ReadUnits},
+		Usage:               &Usage{ReadUnits: derefOrDefault(res.Usage.ReadUnits, 0)},
 		NextPaginationToken: toPaginationToken(res.Pagination),
 	}, nil
 }
@@ -335,7 +336,7 @@ func toUsage(u *data.Usage) *Usage {
 		return nil
 	}
 	return &Usage{
-		ReadUnits: u.ReadUnits,
+		ReadUnits: derefOrDefault(u.ReadUnits, 0),
 	}
 }
 
@@ -372,7 +373,7 @@ func (idx *IndexConnection) akCtx(ctx context.Context) context.Context {
 	newMetadata := []string{}
 	newMetadata = append(newMetadata, "api-key", idx.apiKey)
 
-	for key, value := range idx.additionalMetadata{
+	for key, value := range idx.additionalMetadata {
 		newMetadata = append(newMetadata, key, value)
 	}
 
--- a/pinecone/index_connection.go
+++ b/pinecone/index_connection.go
@@ -75,8 +75,8 @@ func (idx *IndexConnection) UpsertVectors(ctx context.Context, in []*Vector) (ui
 }
 
 type FetchVectorsResponse struct {
-	Vectors map[string]*Vector
-	Usage   *Usage
+	Vectors map[string]*Vector `json:"vectors,omitempty"`
+	Usage   *Usage             `json:"usage,omitempty"`
 }
 
 func (idx *IndexConnection) FetchVectors(ctx context.Context, ids []string) (*FetchVectorsResponse, error) {
@@ -108,9 +108,9 @@ type ListVectorsRequest struct {
 }
 
 type ListVectorsResponse struct {
-	VectorIds           []*string
-	Usage               *Usage
-	NextPaginationToken *string
+	VectorIds           []*string `json:"vector_ids,omitempty"`
+	Usage               *Usage    `json:"usage,omitempty"`
+	NextPaginationToken *string   `json:"next_pagination_token,omitempty"`
 }
 
 func (idx *IndexConnection) ListVectors(ctx context.Context, in *ListVectorsRequest) (*ListVectorsResponse, error) {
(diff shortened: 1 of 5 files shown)


haruska (approved):
:+1:

pinecone/index_connection.go
```diff
@@ -99,9 +99,9 @@ type ListVectorsRequest struct {
 }
 
 type ListVectorsResponse struct {
-	VectorIds           []*string
-	Usage               *Usage
-	NextPaginationToken *string
+	VectorIds           []*string `json:"vector_ids,omitempty"`
```
aulorbe, line 102: Re: your question in the PR description, from what I've read at least it seems like adding `omitempty` is a good call. I think it avoids ambiguous values -- like is a vector ID `0`, or does it not exist? I'm always a fan of being explicit

pinecone/index_connection_test.go
```diff
@@ -243,6 +244,196 @@ func (ts *IndexConnectionTests) TestListVectors() {
...
+					"vec-2": {Id: "vec-2", Values: []float32{0.02, 0.02, 0.02}},
+				},
+				Usage: &Usage{ReadUnits: toUInt32(5)},
+			},
+			want: `{"vectors":{"vec-1":{"id":"vec-1","values":[0.01,0.01,0.01]},"vec-2":{"id":"vec-2","values":[0.02,0.02,0.02]}},"usage":{"read_units":5}}`,
+		},
+		{
+			name:  "Fields omitted",
+			input: FetchVectorsResponse{},
+			want:  `{}`,
```
aulorbe, line 267: Should we account for a case when some fields are present, but others are not (or are empty)?

pinecone/index_connection_test.go
```diff
@@ -243,6 +244,196 @@ func (ts *IndexConnectionTests) TestListVectors() {
...
+		{
+			name: "All fields present",
+			input: QueryVectorsResponse{
+				Matches: []*ScoredVector{
+					{Vector: &Vector{Id: "vec-1", Values: []float32{0.01, 0.01, 0.01}}, Score: 0.1},
+					{Vector: &Vector{Id: "vec-2", Values: []float32{0.02, 0.02, 0.02}}, Score: 0.2},
+				},
+				Usage: &Usage{ReadUnits: toUInt32(5)},
+			},
+			want: `{"matches":[{"vector":{"id":"vec-1","values":[0.01,0.01,0.01]},"score":0.1},{"vector":{"id":"vec-2","values":[0.02,0.02,0.02]},"score":0.2}],"usage":{"read_units":5}}`,
```
aulorbe, line 356: Might be worth at least having the `want` values as globals at the top? Or fixtures or whatever? Just to downsize the amount of duplicated code

pinecone/models.go
```diff
@@ -32,31 +32,31 @@ const (
...
-	Size        *int64
-	Status      CollectionStatus
-	Dimension   *int32
-	VectorCount *int32
-	Environment string
+	Name        string           `json:"name"`
+	Size        *int64           `json:"size,omitempty"`
+	Status      CollectionStatus `json:"status"`
+	Dimension   *int32           `json:"dimension,omitempty"`
+	VectorCount *int32           `json:"vector_count,omitempty"`
```
aulorbe, line 58: This might be an exception to the `omitempty` thing I was saying above -- if `vector_count` is empty, that actually is meaningful, right? Like, that means there are no vectors in the collection, right? If so, I think we should actually _not_ have `omitempty` there, since the value of `0` indicates the actual state of things (vs an ambiguous state)

pinecone/models.go
```diff
@@ -68,47 +68,47 @@ const (
...
+	Region string `json:"region"`
 }
 
 type Vector struct {
-	Id           string
-	Values       []float32
-	SparseValues *SparseValues
-	Metadata     *Metadata
+	Id           string        `json:"id"`
+	Values       []float32     `json:"values,omitempty"`
```
aulorbe, line 91: I thought we always had to have dense values, no? Like, you can technically "only" have sparse values in an index if you upload a dense vector that's just filled with `0`s and has a dimension of `1` or `2` or something, but I think you _have_ to have a dense value (so I think `omitempty` here is not good)

aulorbe: (b/c you should never have an empty dense vector value)

aulorbe (commented):
Some comments

pinecone/index_connection_test.go
austin-denoble: There are cases where we do test for that like in `TestMarshalingDescribeIndexStatsResponse`, but for something like this where all fields have `omitempty` we just test for an empty struct.

pinecone/models.go
austin-denoble: The main reason for using `omitempty` here is `VectorCount` is a pointer, and I've tried to use `omitempty` with all our pointers. The reason is otherwise an empty pointer is marshaled to `null` in the JSON results. 

For collections specifically, this is mainly to cover a specific issue with `VectorCount` and other fields possibly being empty when creating a collection from an empty index.

pinecone/models.go
austin-denoble: Makes sense, I also think empty slices default to `[]` in JSON when marshaling in Go, so that should be fine.

pinecone/index_connection_test.go
austin-denoble: It's definitely a lot of duplicative code. I'd prefer not to move the `want` blocks around though as they're specific to the struct that's being tested, and it'd probably be more confusing to have to look them up elsewhere than in-line.

I can try and think of ways to clean some of this up though.

pinecone/models.go
austin-denoble: You do have to have dense values, yes. Similar to the pointer issue I mentioned here: https://github.com/pinecone-io/go-pinecone/pull/21#discussion_r1591437117 I was mainly doing this defensively to avoid instances where we'd have `values: null` in the JSON, but I don't think that's ever expected to happen.

pinecone/index_connection.go
austin-denoble: I think there are places where not using `omitempty` does make sense as it'll drop zero values that maybe you want to include.

For this PR I was mainly focused on making sure we're using `omitempty` for pointers and structs where it makes sense, primarily to avoid `null` values in the resulting JSON.

pinecone/models.go
austin-denoble: Also the generated code that we map between uses `omitempty` for this, so I think it's fine to leave it as-is for now: https://github.com/pinecone-io/go-pinecone/blob/32391bcb7b862570139e272797ade4ecca50aab3/internal/gen/data/vector_service.pb.go#L88

pinecone/models.go
austin-denoble: I reworked the `Collection` struct and the `toCollection` function to map the pointers into zero values for the ints to make things easier to work with for consumers.

PR #22, opened 2024-05-08
Configure `apiKey`, `additionalHeaders`, and `host` through environment variables

austin-denoble:
## Problem
There was a request to add additional configuration validation to the Go client as you could make requests with an empty `apiKey`. An empty check and early error return were added in a previous PR when passing `apiKey` directly: https://github.com/pinecone-io/go-pinecone/pull/18/files#diff-37c3a14781b4b6e74bcc0cfc8b2462ee2ae243a720a42669256248a3cb005f01R442-R444

We're also lacking the ability to specify an api key or headers through environment variables like the Python client, which can be a nice UX tweak for options in how you configure the client. Additionally, there's no way to specify a control plane host override, which is useful in situations where we need to hit a staging URL.

## Solution
- Refactor `buildClientOptions()` into a method on the `NewClientParams` struct rather than a function which receives the struct.
- Add ability to specify `PINECONE_API_KEY`, `PINECONE_ADDITIONAL_HEADERS`, and `PINECONE_CONTROLLER_HOST` through environment variables. I kept the naming aligned with Python for now.
  - For these values, if the corresponding field has been passed through `NewClientParams` directly, the passed value will override anything set in the environment.
- Add unit tests to validate override behavior, and headers being added as expected. 

## Type of Change

## Test Plan
CI and new unit tests.

---
  - https://app.asana.com/0/0/1207208972408325
  - https://app.asana.com/0/0/1207208968992757

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,3 @@
 PINECONE_API_KEY="<Project API Key>"
-TEST_PINECONE_API_KEY="<Test Project API Key>"
 TEST_POD_INDEX_NAME="<Pod based Index name>"
 TEST_SERVERLESS_INDEX_NAME="<Serverless based Index name>"
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,4 @@
 PINECONE_API_KEY="<Project API Key>"
+TEST_PINECONE_API_KEY="<Test Project API Key>"
 TEST_POD_INDEX_NAME="<Pod based Index name>"
 TEST_SERVERLESS_INDEX_NAME="<Serverless based Index name>"
--- a/.env.example
+++ b/.env.example
@@ -1,3 +1,3 @@
-API_KEY="<Project API Key>"
+PINECONE_API_KEY="<Project API Key>"
 TEST_POD_INDEX_NAME="<Pod based Index name>"
 TEST_SERVERLESS_INDEX_NAME="<Serverless based Index name>"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,4 +21,4 @@ jobs:
         env:
           TEST_POD_INDEX_NAME: ${{ secrets.TEST_POD_INDEX_NAME }}
           TEST_SERVERLESS_INDEX_NAME: ${{ secrets.TEST_SERVERLESS_INDEX_NAME }}
-          TEST_PINECONE_API_KEY: ${{ secrets.API_KEY }}
+          PINECONE_API_KEY: ${{ secrets.API_KEY }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,7 +12,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@v4
         with:
-          go-version: '1.21.x'
+          go-version: "1.21.x"
       - name: Install dependencies
         run: |
           go get ./pinecone
@@ -21,4 +21,4 @@ jobs:
         env:
           TEST_POD_INDEX_NAME: ${{ secrets.TEST_POD_INDEX_NAME }}
           TEST_SERVERLESS_INDEX_NAME: ${{ secrets.TEST_SERVERLESS_INDEX_NAME }}
-          API_KEY: ${{ secrets.API_KEY }}
+          INTEGRATION_PINECONE_API_KEY: ${{ secrets.API_KEY }}
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,4 +21,4 @@ jobs:
         env:
           TEST_POD_INDEX_NAME: ${{ secrets.TEST_POD_INDEX_NAME }}
           TEST_SERVERLESS_INDEX_NAME: ${{ secrets.TEST_SERVERLESS_INDEX_NAME }}
-          INTEGRATION_PINECONE_API_KEY: ${{ secrets.API_KEY }}
+          TEST_PINECONE_API_KEY: ${{ secrets.API_KEY }}
--- a/pinecone/client.go
+++ b/pinecone/client.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"strings"
 
 	"github.com/deepmap/oapi-codegen/v2/pkg/securityprovider"
 	"github.com/pinecone-io/go-pinecone/internal/gen/control"
@@ -24,12 +23,19 @@ type Client struct {
 	sourceTag  string
 }
 
+type NewClientBaseParams struct {
+	Headers    map[string]string
+	Host       string
+	RestClient *http.Client
+	SourceTag  string
+}
+
 type NewClientParams struct {
-	ApiKey     string            // required unless Authorization header provided
-	Headers    map[string]string // optional
-	Host       string            // optional
-	RestClient *http.Client      // optional
-	SourceTag  string            // optional
+	ApiKey     string
+	Headers    map[string]string
+	Host       string
+	RestClient *http.Client
+	SourceTag  string
 }
 
 func NewClient(in NewClientParams) (*Client, error) {
@@ -40,7 +46,7 @@ func NewClient(in NewClientParams) (*Client, error) {
 
 	controlHostOverride := valueOrFallback(in.Host, os.Getenv("PINECONE_CONTROLLER_HOST"))
 	if controlHostOverride != "" {
-		controlHostOverride, err = ensureHTTP(controlHostOverride)
+		controlHostOverride, err = ensureHTTPS(controlHostOverride)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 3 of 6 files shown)


pinecone/client.go
haruska, line 1: Could we have two different constructors? One that requires an ApiKey and the other that requires a different auth mechanism (bearer token? oauth token?). It feels clunky that the "ApiKey is required unless..."

haruska (approved):
Good improvement. It might be more clear to have multiple constructors vs ApiKey `xor` an Authorization header as a part of a headers array.

Technically, the `ApiKey` option is also just setting an Authorization header.

pinecone/client.go
austin-denoble: That does seem reasonable. I'm less clear on what this would look like though since we don't officially offer alternative methods of authentication quite yet. The other clients require an API key for instance.

If the user is able to supply headers directly then they could technically always pass their own auth headers. I think this would also be an issue on Python and TypeScript since they don't do any checking of the headers that are provided. You could also consider this a user-error that they'd need to resolve themselves.

I'm a bit unsure of what would make the most sense here to be honest. 🤔

austin-denoble:
> Good improvement. It might be more clear to have multiple constructors vs ApiKey xor an Authorization header as a part of a headers array.
Technically, the ApiKey option is also just setting an Authorization header.

True, responded here: https://github.com/pinecone-io/go-pinecone/pull/22#discussion_r1596979649

.github/workflows/ci.yaml
```diff
@@ -21,4 +21,4 @@ jobs:
         env:
           TEST_POD_INDEX_NAME: ${{ secrets.TEST_POD_INDEX_NAME }}
           TEST_SERVERLESS_INDEX_NAME: ${{ secrets.TEST_SERVERLESS_INDEX_NAME }}
-          API_KEY: ${{ secrets.API_KEY }}
\ No newline at end of file
+          INTEGRATION_PINECONE_API_KEY: ${{ secrets.API_KEY }}
```
haruska, line 24: This doesn't seem to match up with `.env.example` above

pinecone/client.go
```diff
@@ -16,25 +18,41 @@ import (
...
 
 func NewClient(in NewClientParams) (*Client, error) {
-	clientOptions, err := buildClientOptions(in)
+	clientOptions, err := in.buildClientOptions()
 	if err != nil {
 		return nil, err
 	}
 
-	client, err := control.NewClient("https://api.pinecone.io", clientOptions...)
+	controlHostOverride := valueOrFallback(in.Host, os.Getenv("PINECONE_CONTROLLER_HOST"))
```
haruska, line 47: "CONTROL" vs "CONTROLLER"? (Assuming this is talking about the control plane?

pinecone/client.go
```diff
@@ -16,25 +18,41 @@ import (
 
 type Client struct {
 	apiKey     string
```
haruska, line 20: Is apiKey still needed in the `Client` struct?

pinecone/client.go
```diff
@@ -16,25 +18,41 @@ import (
...
-	Headers    map[string]string // optional
-	RestClient *http.Client      // optional
+	ApiKey     string
+	Headers    map[string]string
+	Host       string
+	RestClient *http.Client
+	SourceTag  string
 }
 
 func NewClient(in NewClientParams) (*Client, error) {
```
haruska, line 41: I'd expect this func to just call `NewClientBase` and configure the ApiKey as an auth header

pinecone/client.go
```diff
@@ -421,39 +460,82 @@ func derefOrDefault[T any](ptr *T, defaultValue T) T {
 	return *ptr
 }
 
-func buildClientOptions(in NewClientParams) ([]control.ClientOption, error) {
+func (ncp *NewClientParams) buildClientOptions() ([]control.ClientOption, error) {
```
haruska, line 463: could buildClientOptions to be a wrapper around buildClientBaseOptions?

pinecone/client_test.go
```diff
@@ -159,13 +185,133 @@ func (ts *ClientTests) TestAuthorizationHeaderOverridesApiKey() {
...
-	authHeaderValue := mockTransport.Req.Header.Get("Authorization")
-	if authHeaderValue != "bearer fooo" {
-		ts.FailNow(fmt.Sprintf("Expected request to have header value 'bearer fooo', but got '%s'", authHeaderValue))
+	testHeaderValue := mockTransport.Req.Header.Get("test-header")
+	assert.Equal(ts.T(), "param-header", testHeaderValue, "Expected request to have header value 'param-header', but got '%s'", testHeaderValue)
+
+	os.Unsetenv("PINECONE_ADDITIONAL_HEADERS")
+}
+
+// func (ts *ClientTests) TestAuthorizationHeaderOverridesApiKey() {
```
haruska, line 194: meant to comment out?

austin-denoble: Meant to remove entirely, the PR wasn't really in a finished state. 😅 The client no longer does any explicit juggling of `ApiKey` and any auth header you may have passed.

pinecone/client.go
austin-denoble: I slimmed things down to just `buildClientBaseOptions`, lemme know if that makes sense.

pinecone/client.go
austin-denoble: It now ultimately calls `NewClientBase` and passes the `ApiKey` through as part of the `Headers` map.

pinecone/client.go
austin-denoble: Removed from `Client` as well as `IndexConnection`.

pinecone/client.go
austin-denoble: I went with `CONTROLLER` because that's the env key that's used in the Python client for the same config: https://github.com/pinecone-io/pinecone-python-client/blob/1fad23f00c114aa8ef33242b91b969da2825c360/pinecone/config/pinecone_config.py#L15

I'm open to changing it up if we want.

.github/workflows/ci.yaml
austin-denoble: I split things out a bit and there are 2 environment keys for setting the API key:
- `PINECONE_API_KEY`
- `INTEGRATION_PINECONE_API_KEY`

Let me know if that's too confusing. Trying to test properly while one env value was used in both situations (the client vs. the tests) felt trickier than just having explicit values for it.

.env.example
```diff
@@ -1,3 +1,4 @@
-API_KEY="<Project API Key>"
+PINECONE_API_KEY="<Project API Key>"
+TEST_PINECONE_API_KEY="<Test Project API Key>"
```
haruska, line 2: Why both?

pinecone/client.go
```diff
@@ -5,44 +5,120 @@ import (
...
+	RestClient *http.Client
+	SourceTag  string
 }
 
 func NewClient(in NewClientParams) (*Client, error) {
-	clientOptions, err := buildClientOptions(in)
-	if err != nil {
-		return nil, err
+	osApiKey := os.Getenv("PINECONE_API_KEY")
+	hasApiKey := (in.ApiKey != "" || osApiKey != "")
```
haruska, line 42: This functionality (of having an environment variable) instead of requiring the `in.ApiKey` seems unclear. On L26 it says that the ApiKey is required. Here, it seems it's optional if I set an environment variable.

pinecone/client.go
```diff
@@ -5,44 +5,120 @@ import (
...
+
+	if !hasApiKey {
+		return nil, fmt.Errorf("no API key provided, please pass an API key for authorization")
+	}
+
+	apiKeyHeader := struct{ Key, Value string }{"Api-Key", valueOrFallback(in.ApiKey, osApiKey)}
```
haruska, line 48: Seems that valueOrFallback could be utilized for the "hasApiKey" check?

pinecone/client.go
```diff
@@ -407,11 +512,25 @@ func decodeCollection(resBody io.ReadCloser) (*Collection, error) {
 	return toCollection(&collectionModel), nil
 }
 
-func minOne(x int32) int32 {
-	if x < 1 {
-		return 1
+func ensureHTTPS(inputURL string) (string, error) {
```
haruska, line 515: can we add the option to disable https? (i'm thinking for local dev)

haruska (approved):
Minor suggestions for your consideration. Good work!

pinecone/client.go
```diff
@@ -5,44 +5,120 @@ import (
...
+	}
+
+	return NewClientBase(NewClientBaseParams{Headers: clientHeaders, Host: in.Host, RestClient: in.RestClient, SourceTag: in.SourceTag})
+}
+
+func NewClientBase(in NewClientBaseParams) (*Client, error) {
+	clientOptions := buildClientBaseOptions(in)
+	var err error
+
+	controlHostOverride := valueOrFallback(in.Host, os.Getenv("PINECONE_CONTROLLER_HOST"))
```
jhamon, line 66: Nice job throwing this in for consistency with other clients.

jhamon (approved):

pinecone/client.go
austin-denoble: The way it's written currently means if you provide a `Host` or `PINECONE_CONTROLLER_HOST`, you're able to use whatever `scheme` as long as `url.Parse()` can read it out. `https` is only applied if there's no scheme, do you think that's sufficient?

For example, you should be able to pass `http://localhost:9999` properly without "https" being added.

haruska: yup that works. The function name is a bit confusing...

pinecone/client.go
austin-denoble: True, I updated the comment to call that out, and also the error that's returned if you don't pass a key. I think adding docstrings we can make this a bit more clear and explicit.

.env.example
austin-denoble: Good point. 😄

Issue #17, opened 2024-04-10
Error responses from Client should include code


haruska:
Without a code included, downstream clients need to parse the error message to understand what the error was.

austin-denoble:
In `v0.5.0` we've added more robust error handling, which should allow easier access to code. See this PR: https://github.com/pinecone-io/go-pinecone/pull/23

PR #23, opened 2024-05-29
Add error handling to REST operations in `Client`

austin-denoble:
## Problem
We're currently not including the code with error responses from `Client`. Additionally, we don't seem to be gracefully handling possible `ErrorResponse` bodies alongside more generic JSON and string responses.

There's an issue tracking this problem here: https://github.com/pinecone-io/go-pinecone/issues/17

Ultimately, we want to surface more useful information to users through errors.

## Solution
This PR focuses on improving the errors that are returned from REST operations in `Client` methods.

- Decode the `http.Response` `Body` once, and then use the `[]byte` value to decode into either errors or response structs.
  - Update `decodeIndex` and `decodeCollection` to take in `[]byte` rather than `io.ReadCloser`. Most of this was done to avoid reading the body to EOF. 
- Add `errorResponseMap` which is used to build up the error information.
- Add `decodeErrorResponse`, `handleErrorResponseBody`, and `formatError` to deal with parsing error payloads and turning them into more readable output.
- Add `PineconeError` which extends the `error` interface and wraps errors in a struct consumers are able to use if they'd like via reflection.

## Type of Change

## Test Plan
Unit tests and a few integration tests have been added to validate `handleErrorResponseBody` and that we're returning `PineconeError` from failed network operations.

---
  - https://app.asana.com/0/0/1207208972408344

### DIFF
--- a/pinecone/client.go
+++ b/pinecone/client.go
@@ -114,17 +114,12 @@ func (c *Client) ListIndexes(ctx context.Context) ([]*Index, error) {
 	}
 	defer res.Body.Close()
 
-	resBodyBytes, err := io.ReadAll(res.Body)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read response body: %w", err)
-	}
-
 	if res.StatusCode != http.StatusOK {
-		return nil, handleErrorResponseBody(resBodyBytes, res.StatusCode, "failed to list indexes: ")
+		return nil, handleErrorResponseBody(res, "failed to list indexes: ")
 	}
 
 	var indexList control.IndexList
-	err = json.Unmarshal(resBodyBytes, &indexList)
+	err = json.NewDecoder(res.Body).Decode(&indexList)
 	if err != nil {
 		return nil, err
 	}
@@ -208,16 +203,11 @@ func (c *Client) CreatePodIndex(ctx context.Context, in *CreatePodIndexRequest) 
 	}
 	defer res.Body.Close()
 
-	resBodyBytes, err := io.ReadAll(res.Body)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read response body: %w", err)
-	}
-
 	if res.StatusCode != http.StatusCreated {
-		return nil, handleErrorResponseBody(resBodyBytes, res.StatusCode, "failed to create index: ")
+		return nil, handleErrorResponseBody(res, "failed to create index: ")
 	}
 
-	return decodeIndex(resBodyBytes)
+	return decodeIndex(res.Body)
 }
 
 type CreateServerlessIndexRequest struct {
@@ -248,16 +238,11 @@ func (c *Client) CreateServerlessIndex(ctx context.Context, in *CreateServerless
 	}
 	defer res.Body.Close()
 
-	resBodyBytes, err := io.ReadAll(res.Body)
-	if err != nil {
-		return nil, fmt.Errorf("failed to read response body: %w", err)
-	}
-
 	if res.StatusCode != http.StatusCreated {
-		return nil, handleErrorResponseBody(resBodyBytes, res.StatusCode, "failed to create index: ")
+		return nil, handleErrorResponseBody(res, "failed to create index: ")
 	}
 
-	return decodeIndex(resBodyBytes)
+	return decodeIndex(res.Body)
 }
 
 func (c *Client) DescribeIndex(ctx context.Context, idxName string) (*Index, error) {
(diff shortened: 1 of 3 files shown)


pinecone/errors.go
```diff
@@ -0,0 +1,12 @@
+package pinecone
+
+import "fmt"
+
+type PineconeError struct {
```
austin-denoble, line 5: This might be unnecessary, but it's somewhat straightforward to adhere to the `error` interface while also adding additional metadata and information that callers can use if they want. Otherwise, the `Error()` output should represent the entirety of the error.

pinecone/client.go
```diff
@@ -463,16 +476,75 @@ func toCollection(cm *control.CollectionModel) *Collection {
...
+	errMap.Body = string(resBodyBytes)
+
+	if errMap.Message != "" {
+		errMap.Message = errMsgPrefix + errMap.Message
+	}
+
+	return formatError(errMap)
+}
+
+func formatError(errMap errorResponseMap) error {
```
austin-denoble, line 538: This is somewhat duplicative and I'm open to changing it. I couldn't really land on a good way of trying to represent things.

This basically just turns the map into a JSON string and stuffs it into the `Msg` in `PineconeError`.

pinecone/client.go
```diff
@@ -463,16 +476,75 @@ func toCollection(cm *control.CollectionModel) *Collection {
...
+
+type errorResponseMap struct {
+	StatusCode int    `json:"status_code"`
+	Body       string `json:"body,omitempty"`
+	ErrorCode  string `json:"error_code,omitempty"`
+	Message    string `json:"message,omitempty"`
+	Details    string `json:"details,omitempty"`
+}
+
+func handleErrorResponseBody(resBodyBytes []byte, statusCode int, errMsgPrefix string) error {
```
austin-denoble, line 511: When testing I noticed that we don't always seem to be guaranteed the `control.ErrorResponse` shape for the body, which would lead to specific errors when trying to decode things.

What I tried to do here was attempt to decode into `ErrorResponse`, and if that fails we just parse the whole body out as a string.

pinecone/client.go
```diff
@@ -107,37 +107,24 @@ func (c *Client) IndexWithAdditionalMetadata(host string, namespace string, addi
...
-}
-
 func (c *Client) ListIndexes(ctx context.Context) ([]*Index, error) {
 	res, err := c.restClient.ListIndexes(ctx)
 	if err != nil {
 		return nil, err
 	}
 	defer res.Body.Close()
 
+	resBodyBytes, err := io.ReadAll(res.Body)
```
austin-denoble, line 117: Ran into EOF problems when trying to read the body in multiple places. To avoid that I went with reading the body initially and then passing `[]byte` around. Let me know if this is inefficient or confusing in some way, maybe there's a better way to be doing this.

pinecone/client_test.go
```diff
@@ -27,6 +29,47 @@ func TestClient(t *testing.T) {
 	suite.Run(t, new(ClientTests))
 }
 
+func TestHandleErrorResponseBody(t *testing.T) {
```
austin-denoble, line 32: These tests are attempting to validate the parsing of various `http.Response` bodies:
- `ErrorResponse`
- generic JSON
- `string`

pinecone/client_test.go
```diff
@@ -285,6 +327,34 @@ func (ts *ClientTests) TestCreatePodIndex() {
 	require.Equal(ts.T(), name, idx.Name, "Index name does not match")
 }
 
+func (ts *ClientTests) TestCreatePodIndexInvalidDimension() {
```
austin-denoble, line 330: I wasn't sure how deep to go with these integration tests that intentionally cause errors. I can add more if they seem useful.

haruska (approved):
Looks good. 

There may be places here where you can avoid reading the entire body up front which might be beneficial in the non-error cases. Those would most likely be cases with large response bodies.

You can probably adjust a lot of the `defer res.Body.Close()` calls and explicitly close after the body read.

These early `io.ReadAll` blocking IO calls may impact performance. If they were minimized to error cases only, it might be better. I'm unsure if that's feasible though.

austin-denoble:
> These early `io.ReadAll` blocking IO calls may impact performance. If they were minimized to error cases only, it might be better. I'm unsure if that's feasible though.

Good point, I refactored things to only decode when we're dealing with an error, or we've got something to return. This is similar to what you had setup before. 👍

PR #26, opened 2024-06-18
Pin oapi-codegen version to latest that uses the /deepmap path

aulorbe:
## Problem

`oapi-codegen` [recently changed its location](https://github.com/oapi-codegen/oapi-codegen?tab=readme-ov-file#action-required-the-repository-for-this-project-has-changed). In order to keep code changes in our `go` repo as minimal as possible, we need to change our `justfile` so that it goes from grabbing the `@latest` release of `oapi-codegen` to grabbing the latest version that _still uses the old path_, which is version `2.2.0`. 

## Solution

Pin version of `oapi-codegen` at `2.2.0`. 

## Type of Change

## Test Plan

CI passes.

---
  - https://app.asana.com/0/0/1207603952846961

### DIFF
--- a/justfile
+++ b/justfile
@@ -7,7 +7,7 @@ test:
 bootstrap:
     go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
     go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@latest
-    go install github.com/deepmap/oapi-codegen/v2/cmd/oapi-codegen@latest
+    go install github.com/deepmap/oapi-codegen/v2/cmd/oapi-codegen@v2.2.0
     go install golang.org/x/tools/cmd/godoc@latest
 gen:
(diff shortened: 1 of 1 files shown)


austin-denoble (approved):
Nice job tracking this one down. This makes sense to me, especially given the link you provided, thanks!

Maybe double check with @haruska as I'd asked him about this dep previously.

haruska (commented):
Looking at the [changelog](https://github.com/oapi-codegen/oapi-codegen/compare/v2.2.0...v2.3.0) seems the diff from v2.2 and v2.3 is largely text changes along with the repo change.

Could we update to the new repo and v2.3 instead of pinning to the old repo?

Edit: Something like this: https://github.com/pinecone-io/go-pinecone/compare/pin_codegen_versions?expand=1

aulorbe:
> Looking at the [changelog](https://github.com/oapi-codegen/oapi-codegen/compare/v2.2.0...v2.3.0) seems the diff from v2.2 and v2.3 is largely text changes along with the repo change.
> 
> Could we update to the new repo and v2.3 instead of pinning to the old repo?
> 
> Edit: Something like this: https://github.com/pinecone-io/go-pinecone/compare/pin_codegen_versions?expand=1

Nice! Closing this PR then in favor of the one you started :)

PR #34, opened 2024-07-01
Add ConfigureIndex method

aulorbe:
## Problem

To have parity with our other clients, we need to include a `ConfigureIndex` method on the Pinecone client. 

[Asana ticket](https://app.asana.com/0/1203260648987893/1207680319590573/f).

## Solution

Add  a `ConfigureIndex` method.  Users can pass a pod size, a # of replicas, or both.

## Type of Change

## Test Plan

New integration tests confirm functionality + pass in CI.


pinecone/client_test.go
```diff
@@ -496,6 +500,145 @@ func (ts *ClientTests) TestDeleteCollection() {
...
+		log.Fatalf("Error creating index %s: %v", ts.configureIndex, erdr)
+	}
+
+	pods := "p1.x1" // test index originally created with "p1.x2" pods
+	replicas := int32(1) // could be nil, but do not want to test nil case here
+	_, err := ts.client.ConfigureIndex(context.Background(), ts.configureIndex, &pods, &replicas)
+	require.ErrorContainsf(ts.T(), err, "Cannot scale down", err.Error())
+
+	deleteErr2 := ts.client.DeleteIndex(context.Background(), ts.configureIndex)
+	if deleteErr != nil {
```
aulorbe, line 527: `Should be deleteErr2, but you get the point`

pinecone/client.go
```diff
@@ -273,6 +273,104 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+//
+//  // To scale both the size of your pods and the number of replicas:
+//   _, err := pc.ConfigureIndex(ctx, "my-index", "p1.x4", 4)
+//   if err != nil {
+//       fmt.Printf("Failed to configure index: %v\n", err)
+//   }
+//
+// [scale a pods-based index]: https://docs.pinecone.io/guides/indexes/configure-pod-based-indexes
+// [app.pinecone.io]: https://app.pinecone.io
+func (c *Client) ConfigureIndex(ctx context.Context, name string, pods *string,
```
austin-denoble, line 315: `pods` may be a little confusing given that fields means something else when creating an index. Although I'm realizing now we don't provide a `pods` value in `CreatePodIndexRequest`.

Still, we use `PodType` there and I think we should do the same here. I know it's a positional arg, but just for clarity in the code here. I know we then assign things to `podType` further down, but initially when I was reading the function I was a bit confused.

pinecone/client.go
```diff
@@ -273,6 +273,104 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+
+	req, err := c.restClient.ConfigureIndex(ctx, name, request)
+	if err != nil {
+		log.Fatalf("Failed to configure index %s. Error: %v", name, err)
+	}
+	if req.StatusCode != http.StatusOK {
+		err := handleErrorResponseBody(req, "failed to configure index: ")
+		if err != nil {
+			return nil, err
+		}
```
austin-denoble, line 360: I think here you can just do:
```suggestion
	return nil, handleErrorResponseBody(req, "failed to configure index: ")
```

We can assume if the `StatusCode` is not `http.StatusOK` we're returning an error.

pinecone/client.go
```diff
@@ -273,6 +273,104 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+				PodType  *control.PodSpecPodType  `json:"pod_type,omitempty"`
+				Replicas *control.PodSpecReplicas `json:"replicas,omitempty"`
+			}{
+				PodType:  podType,
+				Replicas: replicasAmt,
+			},
+		},
+	}
+
+	req, err := c.restClient.ConfigureIndex(ctx, name, request)
```
austin-denoble, line 352: Can we rename `req` to `res` to make it clearer it's the `http.Response` from the function call here?

pinecone/client.go
```diff
@@ -273,6 +273,104 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+		}
+	}
+
+	response, err := control.ParseConfigureIndexResponse(req) // TODO: need this?
+	if err != nil {
+		log.Fatalf("Failed to configure index %s. Error: %v", name, err)
+		return nil, err
+	}
+
+	defer req.Body.Close()
```
austin-denoble, line 369: nit: In other functions we put this line right after the function call so it's a bit more clear we're actually closing it, so I'd maybe move it up to after the initial function call and error check:

```
req, err := c.restClient.ConfigureIndex(ctx, name, request)
if err != nil {
		log.Fatalf("Failed to configure index %s. Error: %v", name, err)
}
defer req.Body.Close()
```

pinecone/client_test.go
```diff
@@ -3,18 +3,18 @@ package pinecone
...
 	"os"
 	"reflect"
 	"strings"
 	"testing"
-
-	"github.com/google/uuid"
-	"github.com/pinecone-io/go-pinecone/internal/mocks"
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
-	"github.com/stretchr/testify/suite"
```
austin-denoble, line 17: nit: I think `goimports` should split these apart automatically right?

pinecone/client.go
```diff
@@ -273,6 +273,104 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+		log.Fatalf("Failed to configure index %s. Error: %v", name, err)
+	}
+	if req.StatusCode != http.StatusOK {
+		err := handleErrorResponseBody(req, "failed to configure index: ")
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	response, err := control.ParseConfigureIndexResponse(req) // TODO: need this?
```
austin-denoble, line 363: I don't think we should be using the internals, I don't think we're doing that anywhere else. I'd maybe look at how we're handling things in something like `DescribeIndex`. Could you juse use the `decodeIndex` function instead?

```suggestion
	res, err := c.restClient.ConfigureIndex(ctx, name, request)
		if err != nil {
		log.Fatalf("Failed to configure index %s. Error: %v", name, err)
	}
	defer res.Body.Close()
	
	if req.StatusCode != http.StatusOK {
	    return nil, handleErrorResponseBody(req, "failed to configure index: ")
	}
	
	return decodeIndex(res.Body)
```

Ultimately this just returns an `IndexModel`, there's not really anything else in `ConfigureIndexResponse`, so you'd need to update the return type and maybe doc string here.

austin-denoble (commented):
Nice work, looks pretty good overall. I've got a few comments and some feedback for your consideration.

pinecone/client.go
```diff
@@ -273,6 +273,104 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+//   if err != nil {
+//       fmt.Printf("Failed to configure index: %v\n", err)
+//   }
+//
+//  // To scale both the size of your pods and the number of replicas:
+//   _, err := pc.ConfigureIndex(ctx, "my-index", "p1.x4", 4)
+//   if err != nil {
+//       fmt.Printf("Failed to configure index: %v\n", err)
+//   }
+//
```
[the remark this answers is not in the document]
aulorbe: Ah I didn't -- this was pushed before before our Slack convo about `gofmt`. I'll run it!

pinecone/client.go
```diff
@@ -273,6 +273,104 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
 	return nil
 }
 
+// ConfigureIndex is used to [scale a pods-based index] up or down by changing the size of the pods or the number of
+//replicas.
+//
+// Parameters:
+//  - name: The name of the index to configure.
+//  - pods: (Optional) The pod size to scale the index to (e.g. for a "p1" pod type,
```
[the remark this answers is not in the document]
aulorbe: Yeah, sweet!

pinecone/client.go
aulorbe: Okay, np, I'll change to `podType`! (I hate how we call it `type` b/c technically it's `size`, but yeah best not to confuse ppl who're used to our other SDKs)

pinecone/client.go
aulorbe: oo nice!

pinecone/client.go
aulorbe: Sure!

pinecone/client.go
aulorbe: okay will do!

pinecone/client_test.go
aulorbe: I thinkkk so. This was also pre-our formatting convo, so I'll push up some changes wrt that soon!

pinecone/client.go
aulorbe: Sure yeah good call, I'll do that!

pinecone/client.go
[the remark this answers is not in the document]
aulorbe: Okay done :)

pinecone/client.go
```diff
@@ -759,6 +759,84 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+	}
+
+	res, err := c.restClient.ConfigureIndex(ctx, name, request)
+	if err != nil {
+		return nil, err
+	}
+
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
```
aulorbe, line 833: Changed this from our usual`http.StatusCreated` to `http.StatusOK`, since configuring an index doesn't give status created or whatnot. Just fyi!

austin-denoble (approved):
Looks good, nice work. Thanks for iterating as well! 🚢

Issue #13, opened 2024-03-04
IndexConnection should allow for more configuration of the gRPC client


haruska:
Probably we will need to configure some additional configs here for stuff like timeouts, custom certs, etc. They are poorly tested / documented because they predate me, but the python client has a [number of different things](https://github.com/pinecone-io/pinecone-python-client/blob/main/pinecone/grpc/config.py) that can be configured for grpc.

_Originally posted by @jhamon in https://github.com/pinecone-io/go-pinecone/pull/11#discussion_r1511610741_

anawishnoff:
This is available in the latest release, closing now.

PR #35, opened 2024-07-02
Allow configuration of `gRPC.ClientConn` through `grpc.DialOption`, refactor `Client.Index()`, bump `grpc` dep to 1.64.0

austin-denoble:
## Problem
We currently do not give users many options for configuring the underlying `grpc.ClientConn` that's used by the `VectorServiceClient` for interacting with the data plane. This will be important for allowing configuration of things like timeouts and proxies, and also for allowing us to unit test aspects of the data plane layer.

We have an open issue to allow more configuration of our gRPC implementation: https://github.com/pinecone-io/go-pinecone/issues/13

## Solution
I ended up implementing a few different things here. Ultimately, I feel like these changes move us more in the right direction for allowing configuration of our gRPC service for data plane operations.

I decided on leaving things a bit open-ended for the consumer. We give them the option to pass variadic `grpc.DialOption` arguments when calling `Client.Index(NewIndexConnParams{...})`. This makes providing these options optional, and you can continue calling the function without them as before.

If someone wants to configure the gRPC connection themselves, they have some latitude to do that via the `Host` and `grpc.DialOption` helper functions: https://pkg.go.dev/google.golang.org/grpc@v1.64.0#DialOption. However, since we're also applying some dial options of our own when constructing the client, allowing this flexibility may require more care when merging things in `newIndexConnection`. Specifically looking for feedback on this if people have opinions. Being able to provide these directly made it possible to unit test some of the behavior as well.

- Bump grpc version from `1.62.0` -> `1.64.0`. @aulorbe and I noticed there was an issue with how we were calling `grpc.Dial()` for setting up the `grpc.ClientConn`. If you passed in an invalid `host`, the `Dial` operation would hang without error'ing for some period of time. I think ultimately we should have been using `grpc.DialContext` to configure the timeout, or using `grpc.WithTimeout()`. When looking at fixing this I bumped our grpc dependency, and noticed the `Dial` and `DialContext` methods are under deprecation notice - they want people using `grpc.NewClient`. It felt like it made sense to go ahead and make this change now:
  - `grpc.Dial` - https://pkg.go.dev/google.golang.org/grpc@v1.64.0#Dial
  - `grpc.NewClient` -  https://pkg.go.dev/google.golang.org/grpc@v1.64.0#NewClient
- Refactor `Client` to have only one `Index` method add `NewIndexConnParams` struct for passing to `Index` rather than discrete arguments. It already felt like our functions for establishing an `IndexConnection` were getting a bit bloated, and when I was looking at adding in gRPC configuration options, it felt like we should go ahead and update this before v1.0.0.
- Add unit test to validate metadata in outgoing RPC requests, and unit tests to validate Client applies the appropriate values to `IndexConnection` through `Client.Index`, `along with Client.extractAuthHeader()`.
- Fix formatting in doc comment code blocks in both `client.go` and `index_connection.go`.
- Bump `google.golang.org/grpc` to v1.65.0 to correct snyk security vulnerability: https://security.snyk.io/vuln/SNYK-GOLANG-GOOGLEGOLANGORGGRPCMETADATA-7430177

## Type of Change

## Test Plan
Make sure CI for unit + integration tests is passing. 

I've added unit tests to specifically target gRPC metadata, testing passing a `grpc.DialOption`, and creating an `IndexConnection` via `Client.Index()`.


internal/utils/interceptors.go
```diff
@@ -0,0 +1,38 @@
...
+	"strings"
+	"testing"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
+)
+
+// MetadataInterceptor is a grpc.UnaryClientInterceptor that extracts the gRPC metadata
+// from the outgoing RPC request context so we can assert on it
+func MetadataInterceptor(t *testing.T, expectedMetadata map[string]string) grpc.UnaryClientInterceptor {
```
austin-denoble, line 15: I spent... too long trying to come up with ways we could mock aspects of the gRPC service to test that things like user-agent and metadata are applied. It was quite challenging.

This should work for validating any metadata we're sending with the outgoing RPC. Validating user-agent was more difficult because `grpc.ClientConn` stores it as a private value until the http2 request is actually constructed.

internal/utils/mocks.go
```diff
@@ -1,4 +1,4 @@
-package mocks
+package utils
```
austin-denoble, line 1: Renamed this package and file since I put the grpc interceptor stuff in here also.

haruska (approved):
Looks good. Nice improvement!

internal/utils/mocks.go
aulorbe: idk if we care about this, but it looks like the rest of this file's changes are just linting? Might want to remove (or at least include in this PR which linter you're using locally, so all contributors can use it so we don't go back and forth on linting changes in the future).

pinecone/client.go
```diff
@@ -37,6 +38,21 @@ type NewClientBaseParams struct {
 	SourceTag  string
 }
 
+// NewIndexConnParams holds the parameters for creating an IndexConnection to a Pinecone index.
+//
+// Fields:
+//   - Host: The host URL of the Pinecone index. This is required. To find your host url use the DescribeIndex or ListIndexes methods.
+//     Alternatively, the host is displayed in the Pinecone web console.
+//   - Namespace: Optional index namespace to use for operations. If not provided, the default namespace of "" will be used.
+//   - AdditionalMetdata: Optional additional metdata to be sent with each RPC request.
```
aulorbe, line 47: Nit: it might be worth giving an example of what metadata one would send with an RPC request

pinecone/client.go
```diff
@@ -80,27 +96,68 @@ func NewClientBase(in NewClientBaseParams) (*Client, error) {
...
+//
+//	 clientParams := pinecone.NewClientParams{
+//	     ApiKey:    "YOUR_API_KEY",
+//		    SourceTag: "your_source_identifier", // optional
+//		}
+//
+//	 pc, err := pinecone.NewClient(clientParams)
+//	 if err != nil {
+//	     log.Fatalf("Failed to create Client: %v", err)
+//		} else {
```
aulorbe, line 122: I think this has to be un-indented one tab (2 spaces)

pinecone/client.go
```diff
@@ -80,27 +96,68 @@ func NewClientBase(in NewClientBaseParams) (*Client, error) {
...
+//
+//	 idx, err := pc.DescribeIndex(ctx, "your-index-name")
+//	 if err != nil {
+//	     log.Fatalf("Failed to describe index \"%s\". Error:%s", idx.Name, err)
+//	 } else {
+//	     fmt.Printf("Successfully found the \"%s\" index!\n", idx.Name)
+//	 }
+//
+//	 indexConnParams := pinecone.NewIndexConnParams{
+//		    Host: idx.Host
```
aulorbe, line 134: I think this has to be un-indented 2 spaces

pinecone/client.go
```diff
@@ -80,27 +96,68 @@ func NewClientBase(in NewClientBaseParams) (*Client, error) {
...
+//	 }
+//
+//	 idxConnection, err := pc.Index(indexConnParams)
+//	 if err != nil {
+//	     log.Fatalf("Failed to create IndexConnection for Host: %v. Error: %v", idx.Host, err)
+//	 } else {
+//		    log.Println("IndexConnection created successfully!")
```
aulorbe, line 141: I think this has to be un-indented 2 spaces

aulorbe (approved):
Yayyy I'm so happy we have just 1 `Index` method now :) Just nits to improve readability :). Nice job!!

internal/utils/mocks.go
austin-denoble: I am using the Go plugin for vscode with the default linting settings. I'm using `staticcheck`, it looks like this: 

Screenshot 2024-07-08 at 1 29 42 PM

Tbh I'm not clear why it introduced a few linting changes here. I wrote the file originally and I don't think it has been changed since. I may not have had format-on-save enabled when I originally wrote it.

PR #36, opened 2024-07-09
Add unit tests for client.go

aulorbe:
## Problem

We do not have unit tests for `client.go`. In an effort to work towards v1.0 readiness, we need them.

## Solution

Add unit tests for key functions/methods in `client.go`, as outlined in [this Asana ticket](https://app.asana.com/0/1203260648987893/1207680319590576/f).

Add commands to `Justfile` to allow contributors to run unit tests _or_ integration tests, if they don't want to run all tests with the usual `just test` cmd.

## Type of Change

## Test Plan

CI passes.


justfile
```diff
@@ -4,6 +4,18 @@ test:
   source .env
   set +o allexport
   go test -count=1 -v ./pinecone
+test-integration:
+    #!/usr/bin/env bash
+    set -o allexport
+    source .env
+    set +o allexport
+    go test -count=1 -v -run TestIntegrationClient ./pinecone
```
aulorbe, line 12: This runs all tests that are methods on the `ClientTestsIntegration` object.

justfile
```diff
@@ -4,6 +4,18 @@ test:
...
+test-unit:
+    #!/usr/bin/env bash
+    set -o allexport
+    source .env
+    set +o allexport
+    go test -v -run Unit ./pinecone
```
aulorbe, line 18: This runs all tests that contain `Unit` in the title.

aulorbe: (Strategy lifted from `Subtests` section of `testing` documentation here:  https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks)

justfile
aulorbe: The idea here is that we can have separate `structs` for integration tests vs. unit tests. Right now, we only have one for integration tests, but we could have one for unit tests in the future. If we do that, we would change the `test-unit` command in the `justfile` to call whatever function we create that actually implements `suite.Run(....)`.

justfile
aulorbe: (See [this comment](https://github.com/pinecone-io/go-pinecone/pull/36/files#r1671376925) for more thoughts.)

justfile
austin-denoble: I know you have another PR for adding tests to index connection as well - is there a way to just use "Integration" for this like the `test-unit` command? It feels like this will only run integration tests covered by `TestIntegrationClient` rather than any that include `Integration`.

pinecone/client_test.go
```diff
@@ -11,14 +12,18 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
+	"github.com/pinecone-io/go-pinecone/internal/gen/control"
+
```
austin-denoble, line 17: nit: These imports seem unordered 🤔 did you run `gofmt`?

pinecone/client_test.go
```diff
@@ -686,11 +581,590 @@ func (ts *ClientTests) TestConfigureIndexHitPodLimit() {
...
+	}
+}
+
+func TestFormatErrorUnit(t *testing.T) {
+	tests := []struct {
+		name     string
+		err      int
+		expected *PineconeError
+	}{
+		{ // TODO: should really add a negative case here, but triggering a marshalling error is hard
```
austin-denoble, line 636: Just curious, what would a negative case be for this?

pinecone/client_test.go
```diff
@@ -686,11 +581,590 @@ func (ts *ClientTests) TestConfigureIndexHitPodLimit() {
...
+		expected string
+	}{
+		{
+			name:     "Confirm https prefix is added",
+			url:      "pinecone-api.io",
+			expected: "https://pinecone-api.io",
+		}, {
+			name:     "Confirm http prefix is added",
+			url:      "http://pinecone-api.io",
+			expected: "http://pinecone-api.io", // TODO: should this be https?
```
austin-denoble, line 766: Nope, it's expected that this should be http if they explicitly provided it. This is helpful for developing against something like Pinecone local.

pinecone/client_test.go
```diff
@@ -686,11 +581,590 @@ func (ts *ClientTests) TestConfigureIndexHitPodLimit() {
...
+				assert.Equal(t, tc.expectedHeaders, client.headers, "Expected headers to be '%v', but got '%v'", tc.expectedHeaders, client.headers)
+			}
+		})
+	}
+}
+
+func TestNewClientBaseUnit(t *testing.T) {
+	// Save the current environment variable value and defer restoring it
+	originalHostEnv := os.Getenv("PINECONE_CONTROLLER_HOST")
+	defer os.Setenv("PINECONE_CONTROLLER_HOST", originalHostEnv)
```
austin-denoble, line 1053: Oh right we can `defer` to do this. I should update some of the other tests...

pinecone/client_test.go
```diff
@@ -686,11 +581,590 @@ func (ts *ClientTests) TestConfigureIndexHitPodLimit() {
...
+			envHost:      "",
+			expectedHost: "",
+			expectedErr:  true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Set environment variable for the test case
+			os.Setenv("PINECONE_CONTROLLER_HOST", tc.envHost)
```
austin-denoble, line 1095: Should this only be set if `tc.envHost != ""`?

austin-denoble (approved):
Nice work! Thanks for splitting out integration and unit tests a bit better, plus adding all this coverage for various things.

Main comment of mine is how we're triggering tests in the justfile and whether that'll work for `index_connection` as well.

pinecone/client_test.go
aulorbe: I did! But I'll do it again :)

justfile
aulorbe: You're right -- for some reason in my head there was a reason I did it differently for integration vs unit tests, but now I can't remember... and it doesn't seem to matter! I'll change :)

pinecone/client_test.go
aulorbe: [we spoke offline & this is actually correct; leaving as-is]

pinecone/client_test.go
aulorbe: uhh idk , ignore this, apologies

pinecone/client_test.go
aulorbe: thanks!

pinecone/client_test.go
aulorbe: no because then the 3rd test fails, since we set the env var `"PINECONE_CONTROLLER_HOST"` to whatever the value is of `tc.envHost`.

For the 3rd test, `tc.envHost` is equal to `""`. So then when we asserts that `tc.expectedHost` is equal to `client.restClient.Server`, it fails because `""` is not equal to `https://api.pinecone.io/`

pinecone/client_test.go
austin-denoble: It's no problem, was just curious

austin-denoble (approved):
LGTM

austin-denoble (approved):

PR #37, opened 2024-07-15
Add docstrings to models.go

aulorbe:
## Problem

There are no docstrings in the `models.go` file.

## Solution

Add docstrings to `models.go`. Since most items are typed structs, follow [these directions](https://go.dev/doc/comment).

## Type of Change

## Test Plan

CI passes.

---
  - https://app.asana.com/0/0/1207689279969119

### DIFF
--- a/pinecone/models.go
+++ b/pinecone/models.go
@@ -8,7 +8,7 @@ import "google.golang.org/protobuf/types/known/structpb"
 type IndexMetric string
 
 const (
-	Cosine     IndexMetric = "cosine"     // Default similarity metric, ideal for textual data
+	Cosine     IndexMetric = "cosine"     // Default distance metric, ideal for textual data
 	Dotproduct IndexMetric = "dotproduct" // Ideal for hybrid search
 	Euclidean  IndexMetric = "euclidean"  // Ideal for distance-based data (e.g. lat/long points)
 )
--- a/pinecone/models.go
+++ b/pinecone/models.go
@@ -2,14 +2,18 @@ package pinecone
 
 import "google.golang.org/protobuf/types/known/structpb"
 
+// IndexMetric is the [distance metric] to be used by similarity search against a Pinecone Index.
+//
+// [distance metric]: https://docs.pinecone.io/guides/indexes/understanding-indexes#distance-metrics
 type IndexMetric string
 
 const (
-	Cosine     IndexMetric = "cosine"
-	Dotproduct IndexMetric = "dotproduct"
-	Euclidean  IndexMetric = "euclidean"
+	Cosine     IndexMetric = "cosine" // Default similarity, ideal for textual data
+	Dotproduct IndexMetric = "dotproduct" // Ideal for hybrid search
+	Euclidean  IndexMetric = "euclidean"  // Ideal for distance-based data (e.g. lat/long points)
 )
 
+// IndexStatusState is the state of a Pinecone Index.
 type IndexStatusState string
 
 const (
@@ -23,6 +27,9 @@ const (
 	Terminating          IndexStatusState = "Terminating"
 )
 
+// Cloud is the [cloud provider] to be used for a Pinecone serverless Index.
+//
+// [cloud provider]: https://docs.pinecone.io/troubleshooting/available-cloud-regions
 type Cloud string
 
 const (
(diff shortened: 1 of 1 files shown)


pinecone/models.go
```diff
@@ -2,14 +2,18 @@ package pinecone
 
 import "google.golang.org/protobuf/types/known/structpb"
 
+// IndexMetric is the [distance metric] to be used by similarity search against a Pinecone Index.
```
aulorbe, line 5: Is the wording awkward here?

austin-denoble: Seems good to me.

pinecone/models.go
```diff
@@ -2,14 +2,18 @@ package pinecone
...
+//
+// [distance metric]: https://docs.pinecone.io/guides/indexes/understanding-indexes#distance-metrics
 type IndexMetric string
 
 const (
-	Cosine     IndexMetric = "cosine"
-	Dotproduct IndexMetric = "dotproduct"
-	Euclidean  IndexMetric = "euclidean"
+	Cosine     IndexMetric = "cosine" // Default similarity, ideal for textual data
```
austin-denoble, line 11: I would use "metric" instead of "similarity".

pinecone/models.go
```diff
@@ -31,16 +38,19 @@ const (
...
 	State IndexStatusState `json:"state"`
 }
 
+// IndexSpec is the infrastructure specification (pods vs serverless) of a Pinecone Index.
 type IndexSpec struct {
 	Pod        *PodSpec        `json:"pod,omitempty"`
 	Serverless *ServerlessSpec `json:"serverless,omitempty"`
 }
 
+// Index is a Pinecone Index object. Can be either a pod-based or a serverless Index, depending on passed IndexSpec.
```
austin-denoble, line 53: nitpick
```suggestion
// Index is a Pinecone Index object. Can be either a pod-based or a serverless Index depending on the IndexSpec.
```

pinecone/models.go
```diff
@@ -67,10 +81,12 @@ const (
 	CollectionStatusTerminating  CollectionStatus = "Terminating"
 )
 
+// PodSpecMetadataConfig is the metadata fields to be indexed when a Pinecone Index is created.
```
austin-denoble, line 84: nit: should it be "are" here?

```suggestion
// PodSpecMetadataConfig are the metadata fields to be indexed when a Pinecone Index is created.
```

pinecone/models.go
```diff
@@ -81,35 +97,57 @@ type PodSpec struct {
...
+// Usage is the usage stats ([Read Units]) for a Pinecone Index.
+//
+// [Read Units]: https://docs.pinecone.io/guides/organizations/manage-cost/understanding-cost#serverless-indexes
 type Usage struct {
 	ReadUnits uint32 `json:"read_units"`
 }
 
+// Filter is the structure that holds the conditions for a query sent to a Pinecone Index,
+// e.g. 'give me all vectors where "author" is "John"'.
 type Filter = structpb.Struct
```
austin-denoble, line 147: I think this type and the `Metadata` type may need clarification.

`Filter` is used primarily in Querying, Deleting, and Describing Index Stats:
- https://github.com/pinecone-io/go-pinecone/blob/cef7a46b73f9e6ce62f4c3eaf7ebb75be548a7c6/pinecone/index_connection.go#L397
- https://github.com/pinecone-io/go-pinecone/blob/cef7a46b73f9e6ce62f4c3eaf7ebb75be548a7c6/pinecone/index_connection.go#L514
- https://github.com/pinecone-io/go-pinecone/blob/cef7a46b73f9e6ce62f4c3eaf7ebb75be548a7c6/pinecone/index_connection.go#L703
- https://github.com/pinecone-io/go-pinecone/blob/cef7a46b73f9e6ce62f4c3eaf7ebb75be548a7c6/pinecone/index_connection.go#L956

`Metadata` is what's used in `Vector`, so upsert and other data plane operations. Now I'm sort of curious why there's two types representing the same thing, maybe we should consolidate? We should do that in a follow-up PR, I don't think you necessarily need to make that change here.

austin-denoble (approved):
Looks good, some nits and then some feedback on `Filter` vs. `Metadata`. It sort of got me wondering whether we should offer both types for what's essentially the same thing. 🤔

pinecone/models.go
aulorbe: Changed to just 'represents'

pinecone/models.go
aulorbe: I think they're different, actually.... 

- In `index_connection`, `Metadata` is the metadata you want to attach to a vector, e.g. in [UpdateVectorRequest](https://github.com/pinecone-io/go-pinecone/blob/main/pinecone/index_connection.go#L776)
- While `Filter` is the metadata filter you want to apply to a request, e.g. in [QueryByVectorValuesRequest](https://github.com/pinecone-io/go-pinecone/blob/main/pinecone/index_connection.go#L397). 

Maybe we're saying the same thing?

aulorbe: [We connected offline; we are going to rename `Filter` to `MetadataFilter` in a follow-up PR.]

PR #38, opened 2024-07-15
Add unit tests for index_connection.go

aulorbe:
## Problem

We do not have unit tests for the private functions inside `index_connection.go`, as outlined in [this Asana ticket](https://app.asana.com/0/1203260648987893/1207762835203548/f).

## Solution

Add 'em! 

Note: I've adjusted the `Justfile` and renamed the integration test components in accordance with the strategy outlined in [this PR](https://github.com/pinecone-io/go-pinecone/pull/36) (unit tests for `client.go`).

## Type of Change

## Test Plan

CI passes.


pinecone/index_connection_test.go
```diff
@@ -555,3 +557,244 @@ func TestMarshalDescribeIndexStatsResponse(t *testing.T) {
...
+		{
+			name: "Pass scored hybrid vector with metadata",
+			scoredVector: &data.ScoredVector{
+				Id:     "hybrid-metadata-1",
+				Values: []float32{0.01, 0.02, 0.03},
+				SparseValues: &data.SparseValues{
+					Indices: []uint32{0, 2},
+					Values:  []float32{0.01, 0.03},
+				},
+				Metadata: &structpb.Struct{
```
aulorbe, line 769: @austin-denoble do you prefer this syntax of constructing the `metadata` obj better than how we do it in [the `toVector` test](https://github.com/pinecone-io/go-pinecone/pull/38/files#diff-6c29b62c08d897a1ab3deb830aea678bf3181775b462b536dc895a0dc3b9bd50R637)? Not sure which is more idiomatic in Go.

pinecone/index_connection_test.go
```diff
@@ -555,3 +559,517 @@ func TestMarshalDescribeIndexStatsResponse(t *testing.T) {
...
+			vector: &data.Vector{
+				Id:     "hybrid-metadata-1",
+				Values: []float32{0.01, 0.02, 0.03},
+				SparseValues: &data.SparseValues{
+					Indices: []uint32{0, 2},
+					Values:  []float32{0.01, 0.03},
+				},
+				Metadata: metadata,
+			},
+			expected: &Vector{
```
austin-denoble, line 641: Nice job covering so many variations here. 👍

pinecone/index_connection_test.go
austin-denoble: I think either are probably fine, although for these I tend to like doing things inline better to co-locate things with the specific test variant. Up to you.

pinecone/index_connection_test.go
```diff
@@ -555,3 +559,517 @@ func TestMarshalDescribeIndexStatsResponse(t *testing.T) {
...
+				Indices: []uint32{0, 2},
+				Values:  []float32{0.01, 0.03},
+			},
+			expected: &data.SparseValues{
+				Indices: []uint32{0, 2},
+				Values:  []float32{0.01, 0.03},
+			},
+		},
+		{
+			name: "Pass sparse values with metadata (metadata is ignored)",
```
austin-denoble, line 926: Why is it ignored / provided if it isn't being used in the test or function being tested? Can we just remove it?

austin-denoble (approved):
These look good, big thanks once again for covering all this testing!

Just a few minor comments, but I don't think there's anything blocking.

pinecone/index_connection_test.go
aulorbe: yeah I think I like inline better too :)

pinecone/index_connection_test.go
aulorbe: I preferred to include it just to account for something weird happening (like in case the user included it, what if somehow it wasn't ignored?), but I can remove it if you want!

austin-denoble (approved):

PR #41, opened 2024-07-16
Add GH action to push Godoc documentation to sdk-docs repo

aulorbe:
## Problem

We need to host our Godoc documentation somewhere via the [sdk-docs repo](https://github.com/pinecone-io/sdk-docs) (rather than localhost), so that users can navigate to `https://sdk.pinecone.io/go` to see documentation.

## Solution

- Add `SSH_DEPLOY_KEY` secret to this repo, retrieved from @austin-denoble, which will allow this repo to push to the `sdk-docs` repo
- Create an GH `action` under new dir `build-docs` that writes an html file (`index.html`) of our documentation to a  new, top-level `docs` dir (`docs` dir not present in public version of Go repo)
- Create a GH `workflow` called `build-and-publish-docs.yml` that should take this `index.html` page and push it to the `sdk-docs` repo, as specified in the `sdk-docs` repo's README.
- Add a package-level synopsis in accordance with `godoc` guidance to `client.go`, so our documentation looks better :)

## Type of Change

## Test Plan

- CI passes
- `index.html` pushed to `sdk-docs`
- Users are able to see documentation at `https://sdk.pinecone.io/go`

---
  - https://app.asana.com/0/0/1207208972408335


.github/actions/build-docs/action.yml
```diff
@@ -0,0 +1,9 @@
+name: 'Build client documentation'
+description: 'Generates client documentation using godoc'
+runs:
+  using: 'composite'
+  steps:
+    - name: Write html file to docs directory
+      run: |
+        godoc -url pkg/github.com/pinecone-io/go-pinecone/pinecone/ > ../../../docs/index.html
```
austin-denoble, line 8: Why does the `/docs/` folder need to be up 3 levels, does it not create it at the root of the project by default?

.github/workflows/build-and-publish-docs.yml
```diff
@@ -0,0 +1,31 @@
...
+
+      - name: Generate Go documentation
+        uses: ./.github/actions/build-docs
+
+      - name: Push documentation artifacts to sdk-docs
+        uses: cpina/github-action-push-to-another-repository@main
+        env:
+          SSH_DEPLOY_KEY: ${{ secrets.SSH_DEPLOY_KEY }}
+        with:
+          source-directory: docs
```
austin-denoble, line 25: This may fail combined with the `../../../docs/index.html` from the action above, I'm not sure though. Is it able to reconcile the folder location?

austin-denoble (approved):
LGTM, one question on the location of the docs output folder.

PR #45, opened 2024-07-17
Try new strategy for better rendering of documentation with gopages

aulorbe:
## Problem

Our current documentation is quite ugly (https://sdk.pinecone.io/go/). We generated it by running `godoc -url`, which basically prints standard html to a file. We then pushed this static file to `sdk-docs` via that repo's README instructions. 

## Solution

There is a Go module called `gopages`: https://johnstarich.com/go/gopages/pkg/github.com/johnstarich/go/gopages/ that seemingly renders things better. We should try that to see if it looks nicer.

## Type of Change

## Test Plan

Need to merge to main, then trigger the documentation workflow in the `sdk-docs` repo to see the rendering.

### DIFF
--- a/.github/workflows/build-and-publish-docs.yml
+++ b/.github/workflows/build-and-publish-docs.yml
@@ -25,7 +25,7 @@ jobs:
         env:
           SSH_DEPLOY_KEY: ${{ secrets.SSH_DEPLOY_KEY }}
         with:
-          source-directory: docs
+          source-directory: dist/pkg/github.com/pinecone-io/go-pinecone/pinecone
           destination-github-username: pinecone-io
           destination-repository-name: sdk-docs
           user-email: clients@pinecone.io
--- a/.github/actions/build-docs/action.yml
+++ b/.github/actions/build-docs/action.yml
@@ -7,15 +7,7 @@ runs:
     - name: Check out the repository
       uses: actions/checkout@v4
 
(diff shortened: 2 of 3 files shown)


.github/workflows/build-and-publish-docs.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         env:
           SSH_DEPLOY_KEY: ${{ secrets.SSH_DEPLOY_KEY }}
         with:
-          source-directory: docs
+          source-directory: dist/pkg/github.com/pinecone-io/go-pinecone/pinecone
```
aulorbe, line 28: `gopages` automatically creates a `dist` dir (and the nested dirs in this path) when run.

aulorbe: the last `pinecone` dir contains a generated `index.html` file, which is what we'll push to `sdk-docs`.

austin-denoble: Interesting, so all the static site content is within that path after running the utility?

austin-denoble (approved):
Seems reasonable, let's see how it goes. 🤞

.github/workflows/build-and-publish-docs.yml
aulorbe: It looks that way on my local, yes!

PR #48, opened 2024-07-19
Centralize Go test suite

aulorbe:
## Problem

Our current test set up is not as useful as it could be. Some issues are:
- The IDs, metadata, and vectors used in tests are generated in non-deterministic ways
- The (integration) tests rely on humans having to manually create indexes in our test project; those indexes must have titles that match Github secret env vars; if this isn't done correctly, tests fail
- The tests do not differentiate between serverless and pod indexes, which make things cumbersome to work with (and make confirming updated vector values, etc. quite impossible)
- Our tests are redundant and cost inefficient -- we spin up indexes mult times and either delete them or just let them live forever in our test project

## Solution

Make it all better! 

The current architecture now looks like this: 

Two top-level test infra files:
1. `integration_test_suite.go`: This file defines a single test `struct` called `IntegrationTests` that holds the fields for everything we need wrt integration testing across all `go` files in our project
    a. Importantly, this file also contains the `testify` mandatory `SetupSuite` and `TeardownSuite` methods attached to this `struct`, so that indexes are always torn down after testing completes 
2. `run_integration_test_suites.go`: This file actually runs the test suites defined in `integration_test_suite.go` via `testify`'s `suite.Run` command. It runs 2 suites: 1 for pods (`podTestSuite`) and 1 for serverless (`serverlessTestSuite`)

Individual test files:
Each file still has a complementary `..._test.go` file that contains its integration and unit tests. The main difference this PR introduces is that each of these files no longer contains a redundant `SetupSuite` function, etc. Instead, they simply call `run_integration_test_suites.go`'s `RunSuites()` method, and everything is automagically created/destroyed.

## Genesis

This refactor arose from Audrey trying to [write integration tests for `update`](https://app.asana.com/0/1203260648987893/1207680319590579/f), but being unable to do so, since she could not easily compare IDs, vector values or metadata before vs after `update` operations.

## Misc.:
There is still a lot of things we can do to make our tests better and more efficient, I'm sure. This is just one baby step on the longer journey towards test suite-maturity.

I don't like it either, but apparently this is what is needed for `testify` to work 😢 . You can't have the `suite.Run` call in the same file as the `SetupSuite` and `TeardownSuite` methods.

> Does the fact that all (integration) tests now share the same struct (`IntegrationTests`) mean that when you run the integration tests in a specific file (e.g. `client_tests.go`), all integration tests actually run?

Yes. This obviously isn't ideal for dev work, but you can figure out how to run individual tests via the command line by reading up on `go test`.

> Why have different `Suites` for pods vs serverless indexes, when they share most fields? 

This is totally fair and tbh I simply didn't refactor this part because this PR is getting gigantic and it seemed like it would add unnecessary complexity to it. But we should go over the pros/cons of having everything in a single `Suite` later! For now, splitting them out produced the invaluable outcome of allowing me to test different things per index type.

 
## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Test Plan

CI passes.


pinecone/index_connection_test.go
```diff
@@ -22,93 +28,121 @@ import (
...
 
 func (ts *IndexConnectionTestsIntegration) SetupSuite() {
+	ctx := context.Background()
+
 	assert.NotEmptyf(ts.T(), ts.host, "HOST env variable not set")
 	assert.NotEmptyf(ts.T(), ts.apiKey, "API_KEY env variable not set")
-	additionalMetadata := map[string]string{"api-key": ts.apiKey}
+	additionalMetadata := map[string]string{"api-key": ts.apiKey, "content-type": "application/json"}
 
-	namespace, err := uuid.NewV7()
```
aulorbe, line 81: Just changed this to make it more up to date w/current Go code I was seeing online

pinecone/index_connection_test.go
```diff
@@ -22,93 +28,121 @@ import (
...
-	assert.NoError(ts.T(), err)
+	namespace, err := uuid.NewUUID()
+	require.NoError(ts.T(), err)
```
aulorbe, line 91: Changed some `assert`s to `require`s since that's what I was seeing is more common with `testify`.

.github/workflows/ci.yaml
```diff
@@ -17,7 +17,8 @@ jobs:
         run: |
           go get ./pinecone
       - name: Run tests
-        run: go test ./pinecone
+        #        run: go test -count=1 -v ./pinecone
```
aulorbe, line 20: We will re-enable this (and remove the line below) once the tests for `client` are working

pinecone/index_connection_test.go
```diff
@@ -373,6 +359,82 @@ func (ts *IndexConnectionTestsIntegration) TestMetadataAppliedToRequests() {
...
+	ctx := context.Background()
+
+	expectedVals := []float32{7.2, 7.2, 7.2, 7.2, 7.2}
+	err := ts.idxConn.UpdateVector(ctx, &UpdateVectorRequest{
+		Id:     ts.vectorIds[0],
+		Values: expectedVals,
+	})
+	assert.NoError(ts.T(), err)
+
+	time.Sleep(5 * time.Second)
```
austin-denoble, line 372: Has 5 seconds worked well for this so far? We had use retries and longer wait windows in some other tests to handle upsert and update. Just curious how it's performed for you so far.

pinecone/index_connection_test.go
```diff
@@ -1073,3 +1135,139 @@ func TestToPaginationToken(t *testing.T) {
...
+	} else {
+		fmt.Printf("Successfully created a new pod index: %s!\n", name)
+	}
+	return podIdx
+}
+
+func generateFloat32Array(n int) []float32 {
+	array := make([]float32, n)
+	for i := 0; i < n; i++ {
+		array[i] = float32(i)
```
austin-denoble, line 1205: This is minor - would be nice to generate random floats or ints for these helpers rather than just using the same index values every time. I guess it ultimately doesn't matter much, but random values could be a nice stress test.

pinecone/index_connection_test.go
```diff
@@ -22,93 +28,119 @@ import (
...
 
-	client, err := NewClient(NewClientParams{ApiKey: apiKey})
-	if err != nil {
-		t.FailNow()
-	}
+	client, err := NewClient(NewClientParams{ApiKey: apiKey, Headers: map[string]string{"content-type": "application/json"}})
+	require.NotNil(t, client, "Client should not be nil after creation")
+	require.NoError(t, err)
 
 	podIndexName := os.Getenv("TEST_PODS_INDEX_NAME")
```
austin-denoble, line 52: Rather than hard-coding this name into an invisible secret in GitHub, I'd prefer if we could generate a random name to use each run since the suite is managing it's resources anyways.

If we hard-code things like this we'd also run into problems if we want to ever run integration tests in parallel, because the tests are no longer isolated and are all pointing at this predefined set of index names. 

Can we look at adding a helper function that generates a random name, possibly with a seed string? Something like we do in our other test suites:
- TypeScript: https://github.com/pinecone-io/pinecone-ts-client/blob/b5a66e6216d7615576dd881a67db627be2aa7bba/src/integration/test-helpers.ts#L92
- Python: https://github.com/pinecone-io/pinecone-python-client/blob/d9df37558ba35b7097dc4bc5c8101b576019caa2/tests/integration/helpers/helpers.py#L13

If we allow the test runs to create their own names and resources and then clean those up, that feels a bit more robust.

austin-denoble (commented):
Took a first pass, nice work getting us onto the path of having our integration tests fully isolated! I think the most important feedback here is around allowing the tests to generate index names, and then rely on those for the test run and cleanup. I'd prefer to move away from storing index names in environment variables if possible.

pinecone/index_connection_test.go
aulorbe: Yes, fantastic idea!

pinecone/index_connection_test.go
aulorbe: Yeah I tried longer and shorter, but 5 was the shortest I found where it still passed

pinecone/index_connection_test.go
aulorbe: Will do!

pinecone/index_connection_test.go
```diff
@@ -22,93 +28,119 @@ import (
...
 
 func TestIntegrationIndexConnection(t *testing.T) {
 	apiKey := os.Getenv("PINECONE_API_KEY")
 	assert.NotEmptyf(t, apiKey, "PINECONE_API_KEY env variable not set")
 
-	client, err := NewClient(NewClientParams{ApiKey: apiKey})
-	if err != nil {
-		t.FailNow()
-	}
+	client, err := NewClient(NewClientParams{ApiKey: apiKey, Headers: map[string]string{"content-type": "application/json"}})
```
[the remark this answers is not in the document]
aulorbe: Yeah it was yelling at me that the content type header was missing, but lemme try again -- maybe that was a red herring for something else at the time

.github/workflows/ci.yaml
aulorbe: Yeah -- lemme rebase your changes (where you added the new env vars for the `client` integration tests) into mine once you merge, and then it'll be done

.github/workflows/ci.yaml
```diff
@@ -17,7 +17,8 @@ jobs:
...
+        run: go test -count=1 -v -run IndexConnection ./pinecone
         env:
           TEST_PODS_INDEX_NAME: ${{ secrets.TEST_PODS_INDEX_NAME }}
           TEST_SERVERLESS_INDEX_NAME: ${{ secrets.TEST_SERVERLESS_INDEX_NAME }}
```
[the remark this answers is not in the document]
aulorbe: Correct

pinecone/index_connection_test.go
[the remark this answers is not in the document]
aulorbe: I've removed and CI still passes... I'm not sure why this is happening but it's not happening any longer.

pinecone/index_connection_test.go
```diff
@@ -1073,3 +1135,139 @@ func TestToPaginationToken(t *testing.T) {
...
+	return uint32(5)
+}
+
+func buildServerlessTestIndex(in *Client, idxName string) *Index {
+	ctx := context.Background()
+
+	fmt.Printf("Creating Serverless index: %s\n", idxName)
+	serverlessIdx, err := in.CreateServerlessIndex(ctx, &CreateServerlessIndexRequest{
+		Name:      idxName,
+		Dimension: int32(setDimensionsForTestIndexes()),
```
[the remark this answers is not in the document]
aulorbe: I just use this new helper function int mult places, so I thought it was better to call a function than to hardcode it to a number

pinecone/index_connection_test.go
aulorbe: I tried doing it by setting the start of the range to a random int, but because `n` is always so low (`5` or less), it often fails. I think we can just leave as is?

austin-denoble: We can leave as-is, it's nitpicky.

I think you could write a helper function that uses `math/rand` maybe:

```go
func RandomFloat32() float32 {
	rand.Seed(time.Now().UnixNano()) // Seed the random number generator
	return rand.Float32()
}
```

.env.example
```diff
@@ -1,3 +1 @@
 PINECONE_API_KEY="<Project API Key>"
-TEST_PODS_INDEX_NAME="<Pod based Index name>"
-TEST_SERVERLESS_INDEX_NAME="<Serverless based Index name>"
```
aulorbe, line 3: buh bai

austin-denoble: 👋

.github/workflows/ci.yaml
```diff
@@ -17,8 +17,6 @@ jobs:
         run: |
           go get ./pinecone
       - name: Run tests
-        run: go test ./pinecone
+        run: go test -count=1 -v ./pinecone
```
austin-denoble, line 20: I think the default for `count` is 1 so you can probably remove explicitly setting it.

README.md
```diff
@@ -101,7 +103,7 @@ Then, execute `just bootstrap` to install the necessary Go packages
 ### .env Setup
 
 To avoid race conditions or having to wait for index creation, the tests require a project with at least one pod index
```
austin-denoble, line 105: nit: I think we can remove this whole first part since it's not longer true that we're avoiding waiting for indexes to create.

pinecone/integration_test_suite.go
```diff
@@ -0,0 +1,218 @@
...
+			break
+		} else {
+			time.Sleep(delay)
+			fmt.Printf("Host \"%s\" not ready for upserting yet, retrying... (%d/%d)\n", ts.host, i, maxRetries)
+		}
+	}
+	return nil
+}
+
+// TODO: how to get this func to work for client tests too
```
austin-denoble, line 123: I think this TODO can come out, right?

pinecone/integration_test_suite.go
```diff
@@ -0,0 +1,218 @@
...
+type IntegrationTests struct {
+	suite.Suite
+	apiKey            string
+	client            *Client
+	host              string
+	dimension         int32
+	indexType         string
+	vectorIds         []string
+	podIdxName        string
+	serverlessIdxName string
```
austin-denoble, line 27: What's the reason for needing both of these names in each struct? It seems like in `run_integration_test_suites.go`, we're creating two different `IntegrationTests` objects. Do they not run separately?

I'm just thinking it feels easier to reason about each instance of the struct handling it's own index. I know you mentioned both serverless and pod index tests run regardless of what file you're trying to test, so there might be something I'm misunderstanding. Basically, my initial thoughts were "why can't this just be `idxName` and each suite handles one index?

pinecone/run_integration_test_suites.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+func RunSuites(t *testing.T) {
```
austin-denoble, line 12: I wonder if there's a way to add an additional parameter here like a `pod` vs. `serverless` enum. We could then directly control what's run when each `client_test.go` or `index_connection_test.go` runs `RunSuites()`, so you could control behavior more directly via integration test file. That feels like it might be nice.

Like you said, there's a lot of ways we could take this and we should probably just take this first step and then play around.

pinecone/index_connection_test.go
```diff
@@ -4,114 +4,26 @@ import (
...
-		additionalMetadata: additionalMetadata,
-		host:               ts.host,
-		namespace:          namespace.String(),
-		sourceTag:          ts.sourceTag})
-	assert.NoError(ts.T(), err)
-	ts.idxConnSourceTag = idxConnSourceTag
-
-	ts.loadData()
+func TestIndexConnectionIntegration(t *testing.T) {
+	RunSuites(t)
```
austin-denoble, line 22: So we call `RunSuites()` from here in each test file (`client_test.go` and `index_connection_test.go`). Then in `RunSuites()` we're creating two `IntegrationTest` structs and then calling `suite.Run` on both of them:

```go
	suite.Run(t, podTestSuite)
	suite.Run(t, serverlessTestSuite)
```

This may be a testify thing I'm not clear on, but is there a possibility we're running duplicates of the test suites? Like if we trigger `go test` on both files, does it spawn different instances of `testing.T` and `RunSuites()` etc?

pinecone/integration_test_suite.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// TODO: how to get this func to work for client tests too
+func GetIndexStatus(ts *IntegrationTests, ctx context.Context) (bool, error) {
+	var indexName string
+	if ts.indexType == "serverless" {
+		indexName = ts.serverlessIdxName
+	} else if ts.indexType == "pods" {
+		indexName = ts.podIdxName
+	}
```
austin-denoble, line 130: I feel like needing to do these `indexType` checks inside of this method is harder to reason about, couldn't we just take in an `indexName` as an argument and then make this function handle one index no matter what?

austin-denoble (approved):
This makes a lot of sense to me, and I really appreciate the clean up here, and putting us in a better direction in terms of integration test robustness. Nice work! 👏 

I do have some questions and follow up around how the `IntegrationTests` suite is created and run. I feel like I'm possibly missing something about our integration testing dependencies. We can follow up offline, or in an additional PR to further refine things. 

For now though we've retained our coverage and better organized the setup and teardown into a centralized location which I love, and is similar to our approach in Java.

pinecone/client_test_2.go
```diff
@@ -0,0 +1 @@
+package pinecone
```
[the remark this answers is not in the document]
aulorbe: (I know this file is renamed, but answering anyways -- apparently no it needs to be there (says my IDE))

.github/workflows/ci.yaml
aulorbe: I thinkkkk from the [documentation](https://pkg.go.dev/cmd/go#hdr-Test_packages) it's not, actually! Check this out: 
Screenshot 2024-07-26 at 12 13 57 PM

pinecone/integration_test_suite.go
aulorbe: Whoopsie yes :)

pinecone/integration_test_suite.go
aulorbe: Yeah....I think you're right! Lemme try it out and see if CI still passes...

pinecone/integration_test_suite.go
aulorbe: gone!

pinecone/index_connection_test.go
aulorbe: Gone! Now we have `RunSuites` in `suite_runner_test.go` (has to be append with `_test` so that `go test` runs the `RunSuites` func, and `test_suite.go`, which is where the setup/teardown lives.

austin-denoble (approved):
Thanks for addressing all that feedback and getting to the root of the test run stuff! 🚢

PR #49, opened 2024-07-22
Add `codegen/build-clients.sh` script, update `codegen/apis` submodule

austin-denoble:
## Problem
We're getting ready to release `v1.0.0` of the Go SDK. Before we do that, we need to make sure we're generating our control and data plane client code from the proper specifications. When the `go-pinecone` repo was originally set up, the `apis` submodule was created and points to a private, Go-specific cut of the OpenAPI and proto files from about 4 months ago.

Since then we're made changes to how we approach client generation, and where our upstream spec files live.

## Solution
- Add new `codegen/` folder to hold both the `apis` submodule and the `build-clients.sh` script.
- Update `.gitmodules` to move `apis` -> `codegen/apis`.
- Add a `build-clients.sh` script. Originally, `just gen` called a few shell commands for generating from OpenAPI and protos. The new script handles generation in a fashion similar to our other repos, and allows specifying a specific API version.
- Update `client.go` / `client_test.go` to handle some changes to underlying types.

My approach was based on the work @jhamon has done in other repos to align our code generation pipelines. 

## Type of Change

## Test Plan
There weren't changes needed to the actual SDK interface. Make sure CI runs and our unit & integration tests are passing as expected.


codegen/build-clients.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
austin-denoble, line 1: I was modeling this after the work @jhamon had done in other repos, but this diverges a bit in that I'm not using a docker image, and we're doing both OAS and proto files in here.

Let me know if this doesn't make sense - or any other feedback. I didn't want to make too many changes to the dependencies we were using to generate code.

codegen/build-clients.sh
```diff
@@ -0,0 +1,70 @@
...
+    popd
+}
+
+verify_spec_version() {
+    local version=$1
+    echo "Verifying spec version has been provided: $version"
+    if [ -z "$version" ]; then
+        echo "Version is required"
+        exit 1
+    fi
```
austin-denoble, line 28: I should maybe do some actual checking against the folder names under `/apis/_build/` maybe. 🤔

codegen/build-clients.sh
```diff
@@ -0,0 +1,70 @@
...
+        exit 1
+    fi 
+}
+
+generate_oas_client() {
+    oas_file="codegen/apis/_build/${version}/control_${version}.oas.yaml"
+
+    oapi-codegen --package=control \
+    --generate types,client \
+    "${oas_file}" > "${control_destination}/control_plane.oas.go"
```
austin-denoble, line 36: There's a `version` in the proto / data plane output, maybe should have that here too.

go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/pinecone-io/go-pinecone
 go 1.21
 
 require (
-	github.com/golang/protobuf v1.5.4
```
austin-denoble, line 6: I ran `go mod tidy` locally and this was removed... I think the `google.golang.org/protobuf v1.34.1` below covers it?

justfile
```diff
@@ -19,11 +19,10 @@ test-unit:
 bootstrap:
     go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.32
     go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3
-    go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@v2.3
+    go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest
     go install golang.org/x/tools/cmd/godoc@latest
 gen:
-  protoc --experimental_allow_proto3_optional --proto_path=apis/proto --go_opt=module="github.com/pinecone-io/go-pinecone" --go-grpc_opt=module="github.com/pinecone-io/go-pinecone" --go_out=. --go-grpc_out=. apis/proto/pinecone/data/v1/vector_service.proto
-  oapi-codegen --package=control --generate types,client apis/openapi/control/v1/control_v1.yaml > internal/gen/control/control_plane.oas.go
+  ./codegen/build-clients.sh "2024-07"
```
austin-denoble, line 25: Calling the script with version from here, we could set a global "API_VERSION" constant somewhere and use that here and elsewhere maybe?

pinecone/client.go
```diff
@@ -498,27 +498,15 @@ func (req CreatePodIndexRequest) TotalCount() *int {
...
-			MetadataConfig *struct {
-				Indexed *[]string `json:"indexed,omitempty"`
-			} `json:"metadata_config,omitempty"`
-			PodType          control.PodSpecPodType   `json:"pod_type"`
-			Pods             *int                     `json:"pods,omitempty"`
-			Replicas         *control.PodSpecReplicas `json:"replicas,omitempty"`
-			Shards           *control.PodSpecShards   `json:"shards,omitempty"`
-			SourceCollection *string                  `json:"source_collection,omitempty"`
-		}{
+	req.Spec = control.IndexSpec{
```
austin-denoble, line 508: Generated anonymous type cleaned up a bit, which is nice.

justfile
```diff
@@ -19,11 +19,10 @@ test-unit:
 bootstrap:
     go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.32
     go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3
-    go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@v2.3
+    go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@latest
     go install golang.org/x/tools/cmd/godoc@latest
 gen:
```
austin-denoble, line 24: `just gen` now requires `just bootstrap` so the script has the right dependencies to call... should it call `bootstrap` itself or is unexpectedly installing dependencies weird?

jhamon (approved):
Looks reasonable to me, assuming tests are passing with the newly built output. Is the version being set in `X-Pinecone-Api-Version` header for control plane actions? For some other languages, I had to augment the generated code by writing the version number somewhere (see [python](https://github.com/pinecone-io/pinecone-python-client/blob/main/codegen/build-oas.sh#L132) and [ts](https://github.com/pinecone-io/pinecone-ts-client/blob/main/codegen/build-oas.sh#L69)) and then setting up that custom header. The generated java included the version number, but still required some effort to hook it up into headers properly. So it's worth double checking that is being done correctly after this refactoring.

.gitignore
```diff
@@ -120,3 +120,4 @@ modules.xml
 
 go.work*
 .env
+.DS_Store
```
haruska, line 123: Should probably just add this to your dev machine as a global gitignore (similar to .idea, .swp files, etc)

go.mod
haruska: Yeah, we should prefer google.golang.org according to https://protobuf.dev/reference/go/faq/.

justfile
haruska: yeah, that would be weird. `bootstrap` shouldn't be a requirement for `just gen`

austin-denoble:
> Is the version being set in X-Pinecone-Api-Version header for control plane actions? For some other languages, I had to augment the generated code by writing the version number somewhere (see [python](https://github.com/pinecone-io/pinecone-python-client/blob/main/codegen/build-oas.sh#L132) and [ts](https://github.com/pinecone-io/pinecone-ts-client/blob/main/codegen/build-oas.sh#L69)) and then setting up that custom header.

Good question, I was going to implement the header in a follow up PR, but given what you mentioned I should probably just do it here. Thanks for the heads up, I had wrongly assumed the version number was written in as part of the spec build process and wasn't considering it.

justfile
```diff
@@ -19,11 +19,10 @@ test-unit:
 bootstrap:
     go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.32
     go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.3
-    go install github.com/oapi-codegen/oapi-codegen/v2/cmd/oapi-codegen@v2.3
```
haruska, line 22: We pinned this to a version to avoid unnecessary codegen updates. It's fine to update but maybe a specific version instead of latest.

justfile
haruska: `just` can also have variables: https://just.systems/man/en/chapter_29.html

pinecone/client.go
```diff
@@ -436,22 +436,22 @@ type CreatePodIndexRequest struct {
 
 // ReplicaCount ensures the replica count of a pods-based Index is >1.
 // It returns a pointer to the number of replicas on a CreatePodIndexRequest object.
-func (req CreatePodIndexRequest) ReplicaCount() *int32 {
+func (req CreatePodIndexRequest) ReplicaCount() int32 {
 	x := minOne(req.Replicas)
-	return &x
+	return x
```
haruska, line 441: is assignment and explicit return necessary?

pinecone/client.go
haruska: This is nice. The previous impl was unsightly.

haruska (approved):
Minor comments. Looks good!

pinecone/client.go
austin-denoble: Nope, lemme clean this up - I ran through here too quick fixing the initial build.

justfile
austin-denoble: Thanks for the catch, I mean to pin to `v2.3.0` which is the version they swapped to the new `github.com` location.

austin-denoble:
> The generated java included the version number, but still required some effort to hook it up into headers properly. So it's worth double checking that is being done correctly after this refactoring.

Added a `generate_version_file()` function to the `build-clients.sh` script. This outputs whichever API version was used to generate the code into a Go file as a constant, which is then imported and used for sending the `X-Pinecone-Api-Version` with requests.

PR #50, opened 2024-07-24
Implement support for `DeletionProtection`

austin-denoble:
## Problem
Our other SDKs support the deletion protection feature with the latest major releases tied to the `2024-07` API version. We need to implement deletion protection in the Go SDK.

## Solution
- Add new `DeletionProtection` enum type to `models.go`.
- Update `CreatePodIndexRequest`, `CreateServerlessIndexRequest`, and `Index` to include `DeletionProtection`.
- Update `ConfigureIndex` to support a `ConfigureIndexParams` struct argument. This aligns more closely with our other SDKs, and makes it easier to pass configurations of arguments without needing to explicitly pass `nil`.

## Type of Change
- [] New feature (non-breaking change which adds functionality)

## Test Plan
Test serverless and pod creation functions to make sure `DeletionProtection` defaults to `disabled`, and vice versa when provided. Test `ConfigureIndex` to make sure configuring an index without passing `DeletionProtection` doesn't override things.

### DIFF
--- a/pinecone/client.go
+++ b/pinecone/client.go
@@ -821,7 +821,7 @@ type ConfigureIndexParams struct {
 //
 // Example:
 //
-//		// To scale the size of your pods from "x2" to "x4":
+//		// To scale the size of your pods-based index from "x2" to "x4":
 //		 _, err := pc.ConfigureIndex(ctx, "my-pod-index", &ConfigureIndexParams{PodType: "p1.x4"})
 //		 if err != nil {
 //		     fmt.Printf("Failed to configure index: %v\n", err)
@@ -1209,14 +1209,14 @@ func toIndex(idx *control.IndexModel) *Index {
 		Ready: idx.Status.Ready,
 		State: IndexStatusState(idx.Status.State),
 	}
-	deletionProtecetion := derefOrDefault(idx.DeletionProtection, control.Disabled)
+	deletionProtection := derefOrDefault(idx.DeletionProtection, "")
 
 	return &Index{
 		Name:               idx.Name,
 		Dimension:          idx.Dimension,
 		Host:               idx.Host,
 		Metric:             IndexMetric(idx.Metric),
-		DeletionProtection: deletionProtecetion,
+		DeletionProtection: DeletionProtection(deletionProtection),
 		Spec:               spec,
 		Status:             status,
 	}
--- a/pinecone/client.go
+++ b/pinecone/client.go
@@ -642,10 +642,6 @@ type CreateServerlessIndexRequest struct {
 //	    }
 func (c *Client) CreateServerlessIndex(ctx context.Context, in *CreateServerlessIndexRequest) (*Index, error) {
 	deletionProtection := toDeletionProtection(in.DeletionProtection)
-
-	fmt.Printf("in.Metric: %v\n", in.Metric)
-	fmt.Printf("control.CreateIndexRequestMetric(in.Metric): %v\n", control.CreateIndexRequestMetric(in.Metric))
-
 	metric := control.CreateIndexRequestMetric(in.Metric)
 
 	req := control.CreateIndexRequest{
@@ -814,9 +810,9 @@ type ConfigureIndexParams struct {
 // [app.pinecone.io]: https://app.pinecone.io
 func (c *Client) ConfigureIndex(ctx context.Context, name string, configParams ConfigureIndexParams) (*Index, error) {
 
-	// if configParams.PodType == "" && configParams.Replicas == 0 && configParams.DeletionProtection == "" {
-	// 	return nil, fmt.Errorf("must specify either PodType, Replicas, or DeletionProtection")
-	// }
+	if configParams.PodType == "" && configParams.Replicas == 0 && configParams.DeletionProtection == "" {
+		return nil, fmt.Errorf("must specify either PodType, Replicas, or DeletionProtection")
+	}
 
 	podType := pointerOrNil(configParams.PodType)
 	replicas := pointerOrNil(configParams.Replicas)
--- a/pinecone/client.go
+++ b/pinecone/client.go
@@ -429,15 +429,16 @@ func (c *Client) ListIndexes(ctx context.Context) ([]*Index, error) {
 //
 // [type of pods]: https://docs.pinecone.io/guides/indexes/choose-a-pod-type-and-size
 type CreatePodIndexRequest struct {
-	Name             string
-	Dimension        int32
-	Metric           IndexMetric
-	Environment      string
-	PodType          string
-	Shards           int32
-	Replicas         int32
-	SourceCollection *string
-	MetadataConfig   *PodSpecMetadataConfig
+	Name               string
+	Dimension          int32
+	Metric             IndexMetric
+	DeletionProtection string
+	Environment        string
+	PodType            string
+	Shards             int32
+	Replicas           int32
+	SourceCollection   *string
+	MetadataConfig     *PodSpecMetadataConfig
 }
 
 // ReplicaCount ensures the replica count of a pods-based Index is >1.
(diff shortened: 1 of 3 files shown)


haruska (approved):

pinecone/client.go
```diff
@@ -548,6 +552,7 @@ func (c *Client) CreatePodIndex(ctx context.Context, in *CreatePodIndexRequest)
 //     and consist only of lower case alphanumeric characters or '-'.
 //   - Dimension: The [dimensionality] of the vectors to be inserted in the Index.
 //   - Metric: The metric used to measure the [similarity] between vectors ('euclidean', 'cosine', or 'dotproduct').
+//   - DeletionProtection:
```
aulorbe, line 555: Is this supposed to be filled in?

pinecone/client.go
```diff
@@ -754,53 +763,99 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
 	return nil
 }
 
+// ConfigureIndexParams contains parameters for configuring an index. For both pod-based
+// and serverless indexes you can configure the DeletionProtection status for an index.
+// For pod-based indexes you can also configure the number of Replicas and the PodType.
+// Each of the fields are optional, but at least one field must be set.
```
aulorbe, line 769: When you say "each of the fields" here, do you mean btwn `replicas` and `podtype` or btwn those 2 + also `deletion protection`?

pinecone/client.go
```diff
@@ -754,53 +763,99 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+// See [scale a pods-based index] for more information.
+//
+// Fields:
+//   - PodType: (Optional) The pod size to scale the index to. For a "p1" pod type,
+//     you could pass "p1.x2" to scale your index to the "x2" size, or you could pass "p1.x4"
+//     to scale your index to the "x4" size, and so forth. Only applies to pod-based indexes.
+//   - Replicas: (Optional) The number of replicas to scale the index to. This is capped by
+//     the maximum number of replicas allowed in your Pinecone project. To configure this number,
+//     go to [app.pinecone.io], select your project, and configure the maximum number of pods.
+//   - DeletionProtection: (Optional) // DeletionProtection determines whether [deletion protection]
```
aulorbe, line 779: Should the sentence here start with `//`?

aulorbe: Might be nice to also hyperlink this to documentation

pinecone/client.go
```diff
@@ -754,53 +763,99 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+// [scale a pods-based index]: https://docs.pinecone.io/guides/indexes/configure-pod-based-indexes
+type ConfigureIndexParams struct {
+	PodType            string
+	Replicas           int32
+	DeletionProtection control.DeletionProtection
+}
+
 // ConfigureIndex is used to [scale a pods-based index] up or down by changing the size of the pods or the number of
-// replicas.
+// replicas, or to enable and disable deletion protection for an index.
```
aulorbe, line 809: Sorry for being behind the times, but so now you can call `ConfigureIndex` on a serverless index?

pinecone/client.go
```diff
@@ -754,53 +763,99 @@ func (c *Client) DeleteIndex(ctx context.Context, idxName string) error {
...
+//		 _, err := pc.ConfigureIndex(ctx, "my-index", nil, nil, &ConfigureIndexParams{DeletionProtection: "enabled"})
+//		 if err != nil {
+//		     fmt.Printf("Failed to configure index: %v\n", err)
+//		 }
 //
 // [scale a pods-based index]: https://docs.pinecone.io/guides/indexes/configure-pod-based-indexes
-// [app.pinecone.io]: https://app.pinecone.io
-func (c *Client) ConfigureIndex(ctx context.Context, name string, podType *string,
-	replicas *int32) (*Index, error) {
+func (c *Client) ConfigureIndex(ctx context.Context, name string, in *ConfigureIndexParams) (*Index, error) {
```
aulorbe, line 849: I like passing in a params struct here, nice thinking

pinecone/client.go
```diff
@@ -1153,13 +1209,16 @@ func toIndex(idx *control.IndexModel) *Index {
 		Ready: idx.Status.Ready,
 		State: IndexStatusState(idx.Status.State),
 	}
+	deletionProtection := derefOrDefault(idx.DeletionProtection, "")
```
aulorbe, line 1212: Should this default to `false` instead of `""`, since

pinecone/client_test.go
```diff
@@ -555,8 +553,8 @@ func (ts *ClientTestsIntegration) TestConfigureIndexIllegalNoPodsOrReplicas() {
 		log.Fatalf("Error creating index %s: %v", name, err)
 	}
 
-	_, err = ts.client.ConfigureIndex(context.Background(), name, nil, nil)
-	require.ErrorContainsf(ts.T(), err, "must specify either podType or replicas", err.Error())
+	_, err = ts.client.ConfigureIndex(context.Background(), name, &ConfigureIndexParams{})
+	require.ErrorContainsf(ts.T(), err, "must specify either PodType, Replicas, or DeletionProtection", err.Error())
```
aulorbe, line 557: I might just be accidentally overlooking it here, but should there also be a test simply testing deletion protection on its own? Like, if you pass in control.ENABLED, and then try to delete an index, you should not be able to, right?

aulorbe (changes requested):
Generally, this looks good to me! I like making the params into a struct and then passing that. 

But I think you might want to add more tests for deletion protection. I see just 1 addition to a test that's already there (the one where you don't pass replicas, podType, or deletionProtection in), but I don't think that's enough... is it possible to add more, like I say [here](https://github.com/pinecone-io/go-pinecone/pull/50#discussion_r1692268414)? (LMK if I'm just missing something and there are more tests!)

pinecone/client.go
aulorbe: (Should we add a test that tries to call `configureIndex` on a serverless index, but passes in `replicas` and `pods` (which obvi aren't things on serverless indexes), so that the test should expect a failure? I think we should!

pinecone/client.go
austin-denoble: Yup, whoops missed updating the struct comments. Thanks!

pinecone/client.go
austin-denoble: Between all 3, you can still pass `Replicas` or `PodType` to the server, it'll just respond with an error if it's a serverless index.

We enforce sending at least one argument in the struct on the client-side.

pinecone/client.go
austin-denoble: Ah I meant to add it as a link in the bottom with the references. Nice catch.

pinecone/client.go
austin-denoble: Yes, let me add a couple tests.

austin-denoble:
> But I think you might want to add more tests for deletion protection. I see just 1 addition to a test that's already there (the one where you don't pass replicas, podType, or deletionProtection in), but I don't think that's enough... is it possible to add more, like I say [here](https://github.com/pinecone-io/go-pinecone/pull/50#discussion_r1692268414)? (LMK if I'm just missing something and there are more tests!)
> 
> And more testing thoughts here [#50 (comment)](https://github.com/pinecone-io/go-pinecone/pull/50#discussion_r1692272005)

Good call, tests added. For now they're testing pod and serverless explicitly.

pinecone/client.go
```diff
@@ -1324,6 +1394,14 @@ func valueOrFallback[T comparable](value, fallback T) T {
 	}
 }
 
+func pointerOrNil[T comparable](value T) *T {
```
aulorbe, line 1397: frickin' Go 🙄

pinecone/client_test.go
```diff
@@ -441,12 +442,78 @@ func (ts *IntegrationTests) TestDeleteCollection() {
...
+		Shards:             1,
+		Replicas:           1,
+	})
+	require.NoError(ts.T(), err)
+
+	// validate we cannot delete the index
+	err = ts.client.DeleteIndex(context.Background(), name)
+	require.ErrorContainsf(ts.T(), err, "failed to delete index: Deletion protection is enabled for this index", err.Error())
+
+	// make sure configuring the index without specifying DeletionProtection maintains "enabled" state
```
aulorbe, line 476: Nice!

aulorbe: (I actually didn't know it worked like this, super cool)

aulorbe (approved):
looks great!

PR #52, opened 2024-07-30
Add `Namespace` to List, Query, and Fetch vector responses

austin-denoble:
## Problem
@jseldess noticed that the Go client doesn't return an explicit `Namespace` with data plane operations. This would be helpful for the end user as `Namespace` is configured when a user targets an index, so having the `Namespace` associated with the specific operation would be useful for validating the namespace where the data was retrieved. Additionally, other SDKs such as TypeScript return namespace as a part of the response object.

## Solution
- Add `Namespace` as a field in `FetchVectorsResponse`, `ListVectorsResponse`, and `QueryVectorsResponse`.
- Update associated unit tests, mainly around marshaling of these structs.

We do not use `omitempty` for `Namespace` as the `""` namespace has meaning as the default.

## Type of Change

## Test Plan
Make sure CI unit / integration tests are passing. Make fetch, query, or list operations and confirm the appropriate `Namespace` is returned depending on which index is targeted.

### DIFF
--- a/pinecone/index_connection.go
+++ b/pinecone/index_connection.go
@@ -200,9 +200,11 @@ func (idx *IndexConnection) UpsertVectors(ctx context.Context, in []*Vector) (ui
 // Fields:
 //   - Vectors: The vectors fetched.
 //   - Usage: The usage information for the request.
+//   - Namespace: The namespace from which the vectors were fetched.
 type FetchVectorsResponse struct {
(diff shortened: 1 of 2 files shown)


haruska (approved):

aulorbe (approved):
Nice, thanks @jseldess !

PR #55, opened 2024-07-30
Add error handling for missing required fields in passed structs

aulorbe:
## Problem

Currently, users structs that are missing fields that are actually required on the backend.

## Solution

- Explicitly note required and optional fields in doc comments
- Add error handling for when structs are passed that are missing required fields
- Add integration tests that confirm an error is thrown when structs are missing required fields

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Test Plan

CI passes

---
  - https://app.asana.com/0/0/1207649770516337


pinecone/client.go
```diff
@@ -1332,7 +1343,7 @@ func formatError(errMap errorResponseMap) error {
 }
 
 func buildClientBaseOptions(in NewClientBaseParams) []control.ClientOption {
-	clientOptions := []control.ClientOption{}
+	var clientOptions []control.ClientOption
```
aulorbe, line 1346: Just clean up

pinecone/index_connection.go
```diff
@@ -1084,7 +1099,7 @@ func sparseValToGrpc(sv *SparseValues) *data.SparseValues {
 }
 
 func (idx *IndexConnection) akCtx(ctx context.Context) context.Context {
-	newMetadata := []string{}
+	var newMetadata []string
```
aulorbe, line 1102: Just cleanup

pinecone/client.go
```diff
@@ -115,10 +115,10 @@ type NewClientBaseParams struct {
 // NewIndexConnParams holds the parameters for creating an IndexConnection to a Pinecone index.
 //
 // Fields:
-//   - Host: The host URL of the Pinecone index. This is required. To find your host url use the DescribeIndex or ListIndexes methods.
+//   - Host: (Required) The host URL of the Pinecone index. To find your host url use the DescribeIndex or ListIndexes methods.
 //     Alternatively, the host is displayed in the Pinecone web console.
 //   - Namespace: Optional index namespace to use for operations. If not provided, the default namespace of "" will be used.
-//   - AdditionalMetdata: Optional additional metdata to be sent with each RPC request.
+//   - AdditionalMetadata: Optional additional metadata to be sent with each RPC request.
```
austin-denoble, line 121: Ah good catch, thank you.

pinecone/client.go
```diff
@@ -280,6 +280,10 @@ func (c *Client) Index(in NewIndexConnParams, dialOpts ...grpc.DialOption) (*Ind
 		in.AdditionalMetadata = make(map[string]string)
 	}
 
+	if in.Host == "" {
+		return nil, fmt.Errorf("host is required to create an IndexConnection. Find your host from calling DescribeIndex or via the Pinecone console")
```
austin-denoble, line 284: Can we capitalize `Host` here since that's how the variable is defined on the struct. We could also call out `ListIndexes` since you can see those values through that as well.

pinecone/client.go
```diff
@@ -504,6 +508,10 @@ func (req CreatePodIndexRequest) TotalCount() int {
 //		       fmt.Printf("Successfully created pod index: %s", idx.Name)
 //	    }
 func (c *Client) CreatePodIndex(ctx context.Context, in *CreatePodIndexRequest) (*Index, error) {
+	if in.Name == "" || in.Dimension == 0 || in.Metric == "" || in.Environment == "" || in.PodType == "" {
+		return nil, fmt.Errorf("name, dimension, metric, environment, and podtype must be included in CreatePodIndexRequest")
```
austin-denoble, line 512: I'd capitalize the keys for the variables in this string as well.

pinecone/client.go
austin-denoble: Was this a gofmt thing / are these equivalent?

I read up a bit and it seems like it works because `append` will initialize the slice if it needs to. Basically:

```go
clientOptions := []control.ClientOption{}
// clientOptions is initialized to an empty slice

var clientOptions []control.ClientOption
// clientOptions is nil in this case
```
Since we know we're going to be appending to the slice it feels a bit more correct to maybe leave it as is.

pinecone/client_test.go
```diff
@@ -255,6 +255,14 @@ func (ts *IntegrationTests) TestCreatePodIndex() {
 	require.Equal(ts.T(), name, idx.Name, "Index name does not match")
 }
 
+func (ts *IntegrationTests) TestCreatePodIndexMissingReqdFields() {
```
austin-denoble, line 258: I think you could argue these are unit tests, since they aren't actually making network calls. I think you could update them to be unit tests instead, right?

pinecone/index_connection.go
austin-denoble: Same as my comment on this change previously, these are technically different under the hood (I think), so we need to make sure this doesn't have any impacts. 

Here's an example, I think the empty slice is handled correctly in most cases but I was just curious: https://go.dev/play/p/f67cXMYYxa3

pinecone/index_connection_test.go
```diff
@@ -256,6 +256,13 @@ func (ts *IntegrationTests) TestUpdateVectorValues() {
 	assert.ElementsMatch(ts.T(), expectedVals, actualVals, "Values do not match")
 }
 
+func (ts *IntegrationTests) TestUpdateVectorMissingReqdFields() {
```
austin-denoble, line 259: Same as above, I think this can be a unit tests since we're expecting a client error before making the request.

austin-denoble (approved):
Nice work! Big thank you for updating all those doc comments to have better annotations around what's expected and optional in the argument structs.

I left some nit-leaning comments around the variable initialization changes - I was mainly curious whether it could have any impact.

Also I think tests where we're expecting a client error before making a network request can just be classified as unit tests since they wont actually be making outbound requests.

pinecone/client.go
aulorbe: Ah okay! My IDE keeps telling me that `Errorf` stmts can't start with a capital letter, so I kept 'em lowercased

austin-denoble: Ah you're right, mine does the same I think it's a Go thing. 🤔 

Might have to lead with a different word if that's the case.

pinecone/client.go
aulorbe: I can change it back, np! It was a `gofmt` thing yes :)

austin-denoble (approved):
Thanks for iterating! 🚢

PR #56, opened 2024-08-01
Add `Host` normalization when building `IndexConnection`

austin-denoble:
## Problem
We had an issue that was logged around trying to call `Client.Index()` with a `Host` value including an `http://` or `https://` schema. The control plane does not attach these schemes, but when building an `IndexConnection` a user needs to provide a host and they could make this mistake. `grpc-go` is not expecting a `target` in `NewClient` that includes a scheme.

We could do a bit of normalization to handle the `Host` more thoroughly.

## Solution
- Add `normalizeHost` helper function in `index_connection.go`. This strips out `http://` & `https://` prefixes, and attaches the port `:443` unless the user has provided one directly. We want to leave some flexibility for making RPCs in specific scenarios.
- Add unit tests for `normalizeHost`.

## Type of Change

## Test Plan
CI unit & integration tests should pass as normal.

To test manually, you can try and prepend schemes to the Host when calling `client.Index()`

```Go
indexConn, err := client.Index(pinecone.NewIndexConnParams{Host: fmt.Sprintf("https://%s", index.Host)})
if err != nil {
    panic(err)
}

stats, err := indexConn.DescribeIndexStats(context.Background())
```

### DIFF
--- a/pinecone/index_connection.go
+++ b/pinecone/index_connection.go
@@ -3,8 +3,8 @@ package pinecone
 import (
 	"context"
 	"crypto/tls"
-	"fmt"
 	"log"
+	"strings"
 
 	"github.com/pinecone-io/go-pinecone/internal/gen/data"
 	"github.com/pinecone-io/go-pinecone/internal/useragent"
@@ -36,7 +36,7 @@ type newIndexParameters struct {
 
 func newIndexConnection(in newIndexParameters, dialOpts ...grpc.DialOption) (*IndexConnection, error) {
 	config := &tls.Config{}
-	target := fmt.Sprintf("%s:443", in.host)
+	target := normalizeHost(in.host)
 
 	// configure default gRPC DialOptions
 	grpcOptions := []grpc.DialOption{
@@ -1011,6 +1011,16 @@ func (idx *IndexConnection) delete(ctx context.Context, req *data.DeleteRequest)
 	return err
 }
 
+func (idx *IndexConnection) akCtx(ctx context.Context) context.Context {
+	newMetadata := []string{}
+
+	for key, value := range idx.additionalMetadata {
+		newMetadata = append(newMetadata, key, value)
+	}
+
+	return metadata.AppendToOutgoingContext(ctx, newMetadata...)
+}
+
 func toVector(vector *data.Vector) *Vector {
 	if vector == nil {
 		return nil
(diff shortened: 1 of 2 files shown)


pinecone/index_connection.go
```diff
@@ -1007,6 +1007,16 @@ func (idx *IndexConnection) delete(ctx context.Context, req *data.DeleteRequest)
 	return err
 }
 
+func (idx *IndexConnection) akCtx(ctx context.Context) context.Context {
```
austin-denoble, line 1010: Moving this up to group with the other `IndexConnection` methods.

pinecone/index_connection_test.go
```diff
@@ -991,6 +991,35 @@ func TestToUsageUnit(t *testing.T) {
...
+func TestNormalizeHost(t *testing.T) {
+	tests := []struct {
+		name         string
+		host         string
+		expectedHost string
+	}{
+		{
+			name:         "http:// scheme should be removed",
+			host:         "http://this-is-my-host.io",
+			expectedHost: "this-is-my-host.io:443",
```
haruska, line 1003: This is a bit odd... I would expect a `http://` transport to default to port 80.

However, I also don't see a user specifying http plaintext without a port for an index.

haruska (approved):
There is a bit of an oddity with stripping `http://` from a host then defaulting to port 443 with TLS enabled on the connection.

pinecone/index_connection.go
```diff
@@ -3,8 +3,9 @@ package pinecone
 import (
 	"context"
 	"crypto/tls"
-	"fmt"
```
aulorbe, line 6: Are you sure this is meant to be deleted?

pinecone/index_connection_test.go
```diff
@@ -991,6 +991,39 @@ func TestToUsageUnit(t *testing.T) {
...
+		}, {
+			name:         "https:// scheme should be removed",
+			host:         "https://this-is-my-host.io",
+			expectedHost: "this-is-my-host.io:443",
+		}, {
+			name:         "http:// scheme without a port should be removed",
+			host:         "http://this-is-my-host.io",
+			expectedHost: "this-is-my-host.io:443",
```
aulorbe, line 1011: What is the difference btwn this test and the 1st test? It looks like they both start and end with the same values, no?

pinecone/index_connection_test.go
```diff
@@ -991,6 +991,39 @@ func TestToUsageUnit(t *testing.T) {
...
+		}, {
+			name:         "http:// scheme and port should be maintained",
+			host:         "http://this-is-my-host.io:8080",
+			expectedHost: "http://this-is-my-host.io:8080",
```
aulorbe, line 1015: It'd be nice to add an iteration here that does `https` with a port, too

aulorbe (approved):
Nothing blocking, just some Qs. Great job!

pinecone/index_connection.go
austin-denoble: It was in my original PR because I removed the `target := fmt.Sprintf("%s:443", in.host)` which was the only usage in here previously.

Once rebasing on your changes though we have those `fmt.Errorf` calls with the input struct checking.

pinecone/index_connection_test.go
austin-denoble: Ah nothing, accidentally duplicated it!

pinecone/index_connection_test.go
austin-denoble: Yeah, I think I meant to add that case and kept one the same through copy pasta or something. 😅

PR #58, opened 2024-08-02
Address bug bash feedback

aulorbe:
During a recent [Go Bug Bash](https://www.notion.so/GO-340b612c52e14891b56f30c44e7b4fd1?pvs=4), we received feedback that some code snippets in the README as well as in the doc comments were either incorrect or not as user-friendly as they could be. This PR addresses that feedback (outlined below for clarity): 

1. [Check to see if index exists before creating it](https://app.asana.com/0/1203260648987893/1207937493683809/f) 
--> done for both Serverless and Pods examples in `README` and `client.go`

2. [Make calls to DescribeIndex print out in a more user-friendly way](https://app.asana.com/0/1203260648987893/1207937493683812/f)
-- > Done in both `README` and  `client.go`

3. [ Remove dereference of ConfigureIndex obj from README](https://app.asana.com/0/1203260648987893/1207937493683816/f)

4. [Add sparse values and metadata example(s) to README](https://app.asana.com/0/1203260648987893/1207937493683820/f)
--> Updated Upsert example in README + linked out to sparse-dense documentation for users to see examples of querying by sparse-dense vectors.

5. [Update `Filter` to be `MetadataFilter` in README](https://app.asana.com/0/1203260648987893/1207937493683824/f)

6. [Remove typo in `QueryByVectorId` in README](https://app.asana.com/0/1203260648987893/1207937493683828/f)

## Type of Change

## Test Plan

CI passes

---
  - https://app.asana.com/0/0/1207937493683828
  - https://app.asana.com/0/0/1207937493683820
  - https://app.asana.com/0/0/1207937493683824
  - https://app.asana.com/0/0/1207937493683816
  - https://app.asana.com/0/0/1207937493683812
  - https://app.asana.com/0/0/1207937493683809

### DIFF
--- a/README.md
+++ b/README.md
@@ -722,7 +722,7 @@ func main() {
 	}
 
 	vectorId := "vector-id"
-	res, err := idxConnection.QueryByVectorId(ctx, &pinecone. & pinecone.QueryByVectorIdRequest{
+	res, err := idxConnection.QueryByVectorId(ctx, &pinecone.QueryByVectorIdRequest{
 		VectorId:      vectorId,
 		TopK:          3,
 		IncludeValues: true,
@@ -730,7 +730,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("Error encountered when querying by vector ID `%v`: %v", vectorId, err)
 	} else {
-		fmt.Printf(prettifyStruct(res))
+		fmt.Printf(prettifyStruct(res.Matches))
 	}
 }
 ```
--- a/README.md
+++ b/README.md
@@ -430,7 +430,13 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to describe index \"%s\". Error:%s", idx.Name, err)
 	} else {
-		fmt.Printf("Successfully found the \"%s\" index!\n", idx.Name)
+		desc := fmt.Sprintf("Description: \n  Name: %s\n  Dimension: %d\n  Host: %s\n  Metric: %s\n"+
+			"  DeletionProtection"+
+			": %s\n"+
+			"  Spec: %+v"+
+			"\n  Status: %+v\n",
+			idx.Name, idx.Dimension, idx.Host, idx.Metric, idx.DeletionProtection, idx.Spec, idx.Status)
+		fmt.Println(desc)
 	}
 }
 ```
--- a/README.md
+++ b/README.md
@@ -150,20 +150,8 @@ func main() {
 
 	indexName := "my-serverless-index"
 
-	indexes, err := pc.ListIndexes(ctx)
-	if err != nil {
-		log.Fatalf("Failed to list indexes: %v", err)
-	}
-
-	for _, idx := range indexes {
-		if idx.Name == indexName {
-			fmt.Printf("Index \"%s\" already exists\n", indexName)
-			return
-		}
-	}
-
 	idx, err := pc.CreateServerlessIndex(ctx, &pinecone.CreateServerlessIndexRequest{
-		Name:      &indexName,
+		Name:      indexName,
 		Dimension: 3,
 		Metric:    pinecone.Cosine,
 		Cloud:     pinecone.Aws,
@@ -171,7 +159,7 @@ func main() {
 	})
 
 	if err != nil {
-		log.Fatalf("Failed to create serverless index: %s", idx.Name)
+		log.Fatalf("Failed to create serverless index: %s", indexName)
 	} else {
 		fmt.Printf("Successfully created serverless index: %s", idx.Name)
 	}
@@ -214,20 +202,8 @@ func main() {
 		Indexed: &[]string{"title", "description"},
 	}
 
-	indexes, err := pc.ListIndexes(ctx)
-	if err != nil {
-		log.Fatalf("Failed to list indexes: %v", err)
-	}
-
-	for _, idx := range indexes {
-		if idx.Name == indexName {
-			fmt.Printf("Index \"%s\" already exists\n", indexName)
-			return
-		}
-	}
-
 	idx, err := pc.CreatePodIndex(ctx, &pinecone.CreatePodIndexRequest{
-		Name:           &indexName,
+		Name:           indexName,
 		Dimension:      3,
 		Metric:         pinecone.Cosine,
 		Environment:    "us-west1-gcp",
--- a/README.md
+++ b/README.md
@@ -370,25 +370,25 @@ func main() {
 	}
 
 	// To scale the size of your pods-based index from "x2" to "x4":
-	_, err := pc.ConfigureIndex(ctx, "my-pod-index", &ConfigureIndexParams{PodType: "p1.x4"})
+	_, err := pc.ConfigureIndex(ctx, "my-pod-index", pinecone.ConfigureIndexParams{PodType: "p1.x4"})
 	if err != nil {
 		fmt.Printf("Failed to configure index: %v\n", err)
 	}
 
 	// To scale the number of replicas to 4:
-	_, err := pc.ConfigureIndex(ctx, "my-pod-index", &ConfigureIndexParams{Replicas: 4})
+	_, err := pc.ConfigureIndex(ctx, "my-pod-index", pinecone.ConfigureIndexParams{Replicas: 4})
 	if err != nil {
 		fmt.Printf("Failed to configure index: %v\n", err)
 	}
 
 	// To scale both the size of your pods and the number of replicas:
-	_, err := pc.ConfigureIndex(ctx, "my-pod-index", &ConfigureIndexParams{PodType: "p1.x4", Replicas: 4})
+	_, err := pc.ConfigureIndex(ctx, "my-pod-index", pinecone.ConfigureIndexParams{PodType: "p1.x4", Replicas: 4})
 	if err != nil {
 		fmt.Printf("Failed to configure index: %v\n", err)
 	}
 
 	// To enable deletion protection:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -148,8 +148,22 @@ func main() {
...
+	for _, idx := range indexes {
+		if idx.Name == indexName {
+			fmt.Printf("Index \"%s\" already exists\n", indexName)
+			return
+		}
+	}
+
 	idx, err := pc.CreateServerlessIndex(ctx, &pinecone.CreateServerlessIndexRequest{
-		Name:      "my-serverless-index",
+		Name:      &indexName,
```
austin-denoble, line 166: This doesn't need to be an address-of, it's not a pointer.

README.md
```diff
@@ -194,12 +208,26 @@ func main() {
...
+	for _, idx := range indexes {
+		if idx.Name == indexName {
+			fmt.Printf("Index \"%s\" already exists\n", indexName)
+			return
+		}
+	}
+
 	idx, err := pc.CreatePodIndex(ctx, &pinecone.CreatePodIndexRequest{
-		Name:           "my-pod-index",
+		Name:           &indexName,
```
austin-denoble, line 230: Same as above, you can just pass the variable in, don't need the `&`.

README.md
```diff
@@ -430,7 +458,13 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to describe index \"%s\". Error:%s", idx.Name, err)
 	} else {
-		fmt.Printf("Successfully found the \"%s\" index!\n", idx.Name)
+		desc := fmt.Sprintf("Description: \n  Name: %s\n  Dimension: %d\n  Host: %s\n  Metric: %s\n"+
+			"  DeletionProtection"+
+			": %s\n"+
+			"  Spec: %+v"+
+			"\n  Status: %+v\n",
```
austin-denoble, line 465: nit: Is this just to fix the printing of the pointer addresses? Not sure if I prefer the formatting to the simpler variety, but up to you.

README.md
```diff
@@ -722,15 +782,15 @@ func main() {
 	}
 
 	vectorId := "vector-id"
-	res, err := idxConnection.QueryByVectorId(ctx, &pinecone. & pinecone.QueryByVectorIdRequest{
```
austin-denoble, line 725: whoops 😬

pinecone/client.go
```diff
@@ -412,8 +412,22 @@ func (c *Client) ListIndexes(ctx context.Context) ([]*Index, error) {
...
+//			if err != nil {
+//				log.Fatalf("Failed to list indexes: %v", err)
+//		}
+//
+//		for _, idx := range indexes {
+//			if idx.Name == indexName {
+//				fmt.Printf("Index \"%s\" already exists\n", indexName)
+//				return
+//			}
+//		}
```
austin-denoble, line 427: nit: double check indentation

pinecone/client.go
```diff
@@ -412,8 +412,22 @@ func (c *Client) ListIndexes(ctx context.Context) ([]*Index, error) {
...
+//
 //	    idx, err := pc.CreatePodIndex(ctx, &pinecone.CreatePodIndexRequest{
-//	        Name:        "my-pod-index",
+//	        Name:        &indexName,
```
austin-denoble, line 430: Don't need pointer.

pinecone/client.go
```diff
@@ -493,20 +507,34 @@ func (req CreatePodIndexRequest) TotalCount() int {
...
+//
+//		for _, idx := range indexes {
+//			if idx.Name == indexName {
+//				fmt.Printf("Index \"%s\" already exists\n", indexName)
+//				return
+//			}
+//		}
+//
+//		idx, err := pc.CreatePodIndex(ctx, &pinecone.CreatePodIndexRequest{
+//		    Name:        &indexName,
```
austin-denoble, line 525: no pointer

pinecone/client.go
```diff
@@ -572,33 +600,47 @@ func (c *Client) CreatePodIndex(ctx context.Context, in *CreatePodIndexRequest)
...
+//			       ApiKey:    "YOUR_API_KEY",
+//			       SourceTag: "your_source_identifier", // optional
+//		    }
+//
+//		    pc, err := pinecone.NewClient(clientParams)
+//		    if err != nil {
+//		        log.Fatalf("Failed to create Client: %v", err)
+//		    } else {
+//			       fmt.Println("Successfully created a new Client object!")
+//		    }
```
austin-denoble, line 615: nit: Possible indentation check

pinecone/client.go
```diff
@@ -572,33 +600,47 @@ func (c *Client) CreatePodIndex(ctx context.Context, in *CreatePodIndexRequest)
...
+//
+//	     indexName := "my-serverless-index"
+//
+//			indexes, err := pc.ListIndexes(ctx)
```
austin-denoble, line 619: nit: Indentation again, sorry

pinecone/client.go
```diff
@@ -572,33 +600,47 @@ func (c *Client) CreatePodIndex(ctx context.Context, in *CreatePodIndexRequest)
...
+//
+//			for _, idx := range indexes {
+//				if idx.Name == indexName {
+//					fmt.Printf("Index \"%s\" already exists\n", indexName)
+//					return
+//				}
+//			}
+//
+//		    idx, err := pc.CreateServerlessIndex(ctx, &pinecone.CreateServerlessIndexRequest{
+//		        Name:      &indexName,
```
austin-denoble, line 632: no pointer

pinecone/client.go
```diff
@@ -640,19 +682,33 @@ type CreateServerlessIndexRequest struct {
...
+//
+//		for _, idx := range indexes {
+//			if idx.Name == indexName {
+//				fmt.Printf("Index \"%s\" already exists\n", indexName)
+//				return
+//			}
+//		}
+//
+//	    idx, err := pc.CreateServerlessIndex(ctx, &pinecone.CreateServerlessIndexRequest{
+//		    Name:    &indexName,
```
austin-denoble, line 700: no pointer

austin-denoble (approved):
Thanks so much for getting to all of these! Lots of nits, and the `Name` in arguments needs to just be a regular value. 🎉

README.md
```diff
@@ -148,8 +148,22 @@ func main() {
...
 	}
 
+	indexName := "my-serverless-index"
+
+	indexes, err := pc.ListIndexes(ctx)
+	if err != nil {
+		log.Fatalf("Failed to list indexes: %v", err)
+	}
+
+	for _, idx := range indexes {
```
evantao96, line 158: The returned error `err` for `CreateServerlessIndex` currently has a message for when the index already exists, so I don't think this is necessary. The "bug" was in the if-statement after this. If the index fails to create, it should log `indexName` not `idx.Name` since idx has not been created.

README.md
```diff
@@ -194,12 +208,26 @@ func main() {
...
 	podIndexMetadata := &pinecone.PodSpecMetadataConfig{
 		Indexed: &[]string{"title", "description"},
 	}
 
+	indexes, err := pc.ListIndexes(ctx)
+	if err != nil {
+		log.Fatalf("Failed to list indexes: %v", err)
+	}
+
+	for _, idx := range indexes {
```
evantao96, line 222: Same as above, I think that this is already handled by the error message

README.md
```diff
@@ -430,7 +458,13 @@ func main() {
 	if err != nil {
 		log.Fatalf("Failed to describe index \"%s\". Error:%s", idx.Name, err)
 	} else {
-		fmt.Printf("Successfully found the \"%s\" index!\n", idx.Name)
+		desc := fmt.Sprintf("Description: \n  Name: %s\n  Dimension: %d\n  Host: %s\n  Metric: %s\n"+
+			"  DeletionProtection"+
+			": %s\n"+
+			"  Spec: %+v"+
+			"\n  Status: %+v\n",
+			idx.Name, idx.Dimension, idx.Host, idx.Metric, idx.DeletionProtection, idx.Spec, idx.Status)
```
evantao96, line 466: Should `*idx.Spec` be printed, rather than `idx.Spec`?

Similarly, `*idx.Status` rather than `idx.Status`?

evantao96 (approved):

README.md
austin-denoble: I believe using `%+v` will automatically print the pointer, but I'm actually not positive. You may be right.

README.md
austin-denoble: Yeah, look on n 174, we can't access `idx` if there was an error. Nice catch, Evan.

aulorbe: Ahh thank you! I was misunderstanding your feedback from the bug bash :)

README.md
aulorbe: yeah, it's just some UX lipstick that was brought up in the bug bash. I agree it's sorta ugly, but the printed-out statement is really a lot better. I think we should keep?

README.md
aulorbe: There's no difference (in the output) between printing out `idx.Spec` vs `*idx.Spec`. Right now, the output looks like this (I think drilling down to `Pod` or `Serverless` within `Spec` is too ugly): 

Screenshot 2024-08-02 at 2 13 06 PM

evantao96: The '*' gets rid of the '&' in front of the struct in Spec and Status

README.md
austin-denoble: Sounds good.

austin-denoble (approved):
🚢

evantao96 (approved):

PR #63, opened 2024-08-06
FIx README v1.0.0

austin-denoble:
## Problem
Missed completely removing the old `metadata` example in Query by Vector. Nice catch @jseldess.

## Solution
Fix it.

## Type of Change
- [X] Non-code change (docs, etc)

## Test Plan
N/A


jseldess (approved):
LGTM

PR #66, opened 2024-08-14
Add GH `release` workflow and `bump-version` action, fix hard-coded client version

austin-denoble:
## Problem
We have a private function which returns a hard-coded version number which is used when we build the user-agent for requests originating from the client. This is difficult to maintain while releasing because Go releases are handled through GitHub and Git tags, and requires us to keep things in sync manually.

I also missed updating this for `v1.0.0`, so it needs to be bumped anyways.

We're also missing any kind of "official" release process for the Go client, which would be beneficial in standardizing versioning and the process for releasing the client now that we're >v1.0.0. I'd rather we have a GitHub workflow similar to the other clients even if it primarily involves bumping a hard-coded version file, and pushing a tag.

## Solution
- Replace the old hard-coded version with a new `internal/version.go` file which packages and holds `internal.Version` to be used within the client. This is the file that CI will manage and commit updates towards as we bump versions.
- Add new GitHub action `.github/actions/bump-version` allowing us to easily bump the `currentVersion` using a specific `releaseLevel` (major, minor, patch),  `isPrerelease`, and `prereleaseSuffix` if desired. This action was something @jhamon originally added to the [pinecone-python-client](https://github.com/pinecone-io/pinecone-python-client/tree/main/.github/actions/bump-version). I've lifted it here with minor alterations to support just returning a new `version_tag` rather than modifying a file using `fs` which it does in the Python repo. 
    - I would like to work on getting this action shared properly through GitHub so that we can reuse with one source of truth, but that will take some additional work beyond the scope of this fix. I've added a ticket: https://app.asana.com/0/1203260648987893/1208039955828667/f
- Add new GitHub workflow `./github/workflows/release.yaml` to facilitate releasing the Go SDK by updating `internal/version.go`, and pushing the new tag. For prerelease, we just push the tag without committing.

## Type of Change

## Test 
I'll need to get the CI files into source before I can test the actual release process. My thinking here is I've updated the hard-coded value to the current `v1.0.0`, and I'd like to release a `v1.0.1` to fix the current problem with the user-agent not matching. I think this is reasonable, but let me know if you don't agree.

The `bump-version` action itself was lifted directly from @jhamon's work including unit tests, so I think that is mostly safe. I'd spend the bulk of the review looking at my approach in `release.yaml`.

### DIFF
--- a/.github/actions/bump-version/.gitignore
+++ b/.github/actions/bump-version/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
--- a/.github/actions/bump-version/core.js
+++ b/.github/actions/bump-version/core.js
@@ -1,0 +1,79 @@
+// Copied these commands out of the github actions toolkit
+// because actually depending on @actions/core requires me to check
+// in node_modules and 34MB of dependencies, which I don't want to do.
+
+const fs = require("fs");
+const os = require("os");
+
+function getInput(name, options) {
+  const val =
+    process.env[`INPUT_${name.replace(/ /g, "_").toUpperCase()}`] || "";
+  if (options && options.required && !val) {
+    throw new Error(`Input required and not supplied: ${name}`);
+  }
+
+  if (options && options.trimWhitespace === false) {
+    return val;
+  }
+
+  return val.trim();
+}
+
+function toCommandValue(input) {
+  if (input === null || input === undefined) {
+    return "";
+  } else if (typeof input === "string" || input instanceof String) {
+    return input;
+  }
+  return JSON.stringify(input);
+}
+
+function prepareKeyValueMessage(key, value) {
+  const delimiter = `delimiter_${Math.floor(Math.random() * 100000)}`;
+  const convertedValue = toCommandValue(value);
+
+  // These should realistically never happen, but just in case someone finds a
+  // way to exploit uuid generation let's not allow keys or values that contain
+  // the delimiter.
+  if (key.includes(delimiter)) {
+    throw new Error(
+      `Unexpected input: name should not contain the delimiter "${delimiter}"`
+    );
+  }
+
+  if (convertedValue.includes(delimiter)) {
+    throw new Error(
+      `Unexpected input: value should not contain the delimiter "${delimiter}"`
+    );
+  }
+
+  return `${key}<<${delimiter}${os.EOL}${convertedValue}${os.EOL}${delimiter}`;
+}
+
+function setOutput(name, value) {
+  const filePath = process.env["GITHUB_OUTPUT"] || "";
+  if (filePath) {
+    return issueFileCommand("OUTPUT", prepareKeyValueMessage(name, value));
+  }
+
+  process.stdout.write(os.EOL);
+  issueCommand("set-output", { name }, toCommandValue(value));
+}
+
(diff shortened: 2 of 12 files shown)


.github/actions/bump-version/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules/
```
austin-denoble, line 1: This entire action was lifted from work @jhamon did previously: https://github.com/pinecone-io/pinecone-python-client/pull/178

I'll call out lines I modified, which should be within `action.js` and `index.js` where I got rid of modifying local files via `fs.readFileSync` and `fs.writeFileSync`. I'm doing the file manipulating using `sed` in the `release.yaml` file.

.github/actions/bump-version/index.js
```diff
@@ -0,0 +1,8 @@
+const action = require("./action");
+const core = require("./core");
+
+action.bumpVersion(
+  core.getInput("currentVersion"),
+  core.getInput("bumpType"),
+  core.getInput("prereleaseSuffix")
+);
```
austin-denoble, line 8: Got rid of `fs.readFileSynx`, `fs.writeFileSynx`, and doing anything with the `newVersion` that used to get returned from the `action.bumpVersion` call. `bumpVersion` uses `core.setOutput` and then we use the output in the `release.yaml` workflow.

.github/actions/bump-version/package-lock.json
```diff
@@ -0,0 +1,3770 @@
+{
```
austin-denoble, line 1: A bummer, but if we can move all of this junk into its own public repo and then reuse the action across our repos that would be lovely.

internal/useragent/useragent.go
```diff
@@ -3,11 +3,12 @@ package useragent
...
 	"fmt"
 	"strings"
+
+	"github.com/pinecone-io/go-pinecone/internal"
 )
 
 func getPackageVersion() string {
-	// update at release time
-	return "v0.5.0"
+	return internal.Version
```
austin-denoble, line 11: This is the only fix to the code itself, and we're expecting the `internal.Version` value to be somewhat managed through the CI process.

haruska: nice!

internal/useragent/useragent.go
```diff
@@ -3,11 +3,12 @@ package useragent
 import (
 	"fmt"
 	"strings"
+
+	"github.com/pinecone-io/go-pinecone/internal"
```
haruska, line 7: isn't useragent already in this package?

haruska (approved):
:+1:

jhamon (approved):

.github/actions/bump-version/core.js
```diff
@@ -0,0 +1,79 @@
...
+    return input;
+  }
+  return JSON.stringify(input);
+}
+
+function prepareKeyValueMessage(key, value) {
+  const delimiter = `delimiter_${Math.floor(Math.random() * 100000)}`;
+  const convertedValue = toCommandValue(value);
+
+  // These should realistically never happen, but just in case someone finds a
```
aulorbe, line 35: nice

aulorbe (approved):

PR #70, opened 2024-08-23
Pass `currentVersion` in `release` workflow

austin-denoble:
## Problem
I tried running the new `release` action and it errored when trying to push to `main` and when generating the new version tag due to 3 things:
- `main` branch doesn't allow pushes without PRs, we need to disable that to allow more flexibility from CI (we do this in the python and typescript repos).
- The `currentVersion` retrieved using `git describe` was not being passed to the `bump-version` action, so we ended up with NaN in the output.
- The `prereleaseSuffix` defaulting to "alpha1" and then being passed to `bump-version` was an issue because it was being appended whether or not `isPrerelease` was true.

Failed run: https://github.com/pinecone-io/go-pinecone/actions/runs/10529783970/job/29178282163

## Solution
- Add explicit `SEMVER_VERSION` and `PRERELEASE_SUFFIX` variables to the `release` workflow. Make sure these values are set in the `get_tag_version` and `set_prerelease_suffix` steps.
- Pass the `currentVersion` value to `bump-version` properly. 

## Type of Change

## Test Plan
Push changes up and rerun the version bump.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,12 +26,13 @@ on:
         description: "Suffix to add to version number for marking as a pre-release alpha or beta client. Value ignored when isPrerelease is false"
         required: false
         type: string
-        default: "alpha1"
+        default: ""
 
 jobs:
   bump-version-and-release:
     runs-on: ubuntu-latest
-
+    env:
+      SEMVER_VERSION: "" # Set by the get_tag_version step
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -47,21 +49,27 @@ jobs:
...
+        if: ${{ inputs.isPrerelease == true }}
+        run: |
+          echo "PRERELEASE_SUFFIX=${{ inputs.prereleaseSuffix }}" >> $GITHUB_ENV
 
       - name: Bump version
         id: bump
         uses: "./.github/actions/bump-version"
         with:
-          versionFile: pinecone/__version__
+          currentVersion: ${{ env.SEMVER_VERSION}}
```
austin-denoble, line 70: Copy/pasta mistake from the original workflow I used as a template. The underlying `bump-version` action seems to work as expected and is unit tested, I just wasn't passing the proper input.

jhamon (approved):
👍

aulorbe (approved):

PR #71, opened 2024-09-05
Return error for serverless index creation failure

jseldess:
## Problem

The create serverless index example doesn't return the error on index creation failure. 

## Solution

Fix the error message.

## Type of Change

## Test Plan

Describe specific steps for validating this change.


austin-denoble (approved):
Good catch, thanks Jesse.

PR #72, opened 2024-09-10
Refactor sparse vector integration tests

austin-denoble:
## Problem
We had a ticket logged for making sure that we covered working with sparse vectors in integration tests.

## Solution
I realized we were already creating sparse vectors whenever we called `createVectorsForUpsert`, but the function was a bit unwieldy to work with. 

I refactored things into a new helper `func GenerateVectors(numOfVectors int, dimension int32, isSparse bool) []*Vector`. This is a bit more flexible.

I also cleaned up how we were handling the `IntegrationTest.vectorIds` field on the test struct. I wanted to make sure we were consistently adding to and deleting from this collection across tests.

## Type of Change

## Test Plan
Existing integration tests should pass.

---
  - https://app.asana.com/0/0/1207937493683805

### DIFF
--- a/pinecone/index_connection_test.go
+++ b/pinecone/index_connection_test.go
@@ -66,17 +66,6 @@ func (ts *IntegrationTests) TestDeleteVectorsById() {
 	}
 }
 
-func (ts *IntegrationTests) TestUpsertSparseVectors() {
-	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


haruska (approved):
:+1:

jhamon (approved):

PR #74, opened 2024-09-13
Fix `normalizeHost`, allow for non-secure connections

austin-denoble:
## Problem
When working with specific environments and hosts that require a non-secure connection, calling `pc.Index(pinecone.NewIndexConnParams{Host: idx.Host})` with a `Host` value that is insecure, you'll run into unexpected behavior seeing `:443` attached as a port, or gRPC errors around connecting insecurely without explicitly passing a `grpc.DialOption`.

Our `normalizeHost` function has a bug in that it applies port `:443` to `Host` values which are passed without an explicit scheme.

## Solution
- We can simplify the `normalizeHost` function to only deal with stripping away any provided scheme, and then checking to see if the connection is explicitly insecure. This is accomplished by providing `http://` as a part of the `NewIndexConnParams{ Host: yourHost }` value, which would need to be accounted for in documentation, etc.

This felt most straightforward because of how the `grpc` module handles host addresses. Ultimately, we need to strip the scheme off any hosts that are passed, but we also need to decide whether we're dialing in a secure or insecure fashion. I thought of providing an explicit `bool` as a part of the `NewIndexConnParams`, but it felt more unwieldy adding new fields.

## Type of Change

## Test Plan

I have a local test harness that I'm using to interact with a locally hosted index. I was overriding the `go-pinecone` package in `go.mod` to work against my local copy of the SDK. You can pull this branch down and do something similar locally. Here are some of my results:

### Before
Screenshot 2024-09-12 at 10 09 45 PM

### After
Screenshot 2024-09-12 at 10 33 59 PM

### DIFF
--- a/pinecone/index_connection.go
+++ b/pinecone/index_connection.go
@@ -12,6 +12,7 @@ import (
 	"github.com/pinecone-io/go-pinecone/internal/useragent"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/metadata"
 )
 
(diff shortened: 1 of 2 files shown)


haruska (approved):
:+1:

PR #77, opened 2024-09-18
Add Pinecone Local testing support in CI

austin-denoble:
## Problem
We want to add CI that validates testing the Go SDK against an instance of pclocal hosted in a Docker container.

## Solution
- Update the `ci.yaml` GitHub workflow file to add a new `services` block, which includes using the `ghcr.io/pinecone-io/pinecone-index:latest` Docker image to create and how two indexes locally (pods and serverless).
- Add new `/pinecone/local_test.go` which defines a new `LocalIntegrationTests` struct, which isolates a collection of tests specifically meant for pclocal. The tests are isolated from our unit + integration test suite with the `//go:build localServer` build tag. Two suites are defined for pods and serverless, and each set of tests is run against both indexes. For now the test suite generates 100 vectors, upserts them to the index, tests basic Fetch, Query, Update, DescribeIndexStats, and Delete.
- Add a new `Run local integration tests` step to the `ci.yaml` workflow. This triggers the new `local-test.go` file which tests both instances of locally hosted indexes (`PINECONE_INDEX_URL_POD`, `PINECONE_INDEX_URL_SERVERLESS`) with `go test -count=1 -v ./pinecone -run TestRunLocalIntegrationSuite -tags=localServer`.

### To Do
I'd like to expand the number of different tests we exercise on each index. Specifically:
- Metadata filtering
- Sparse vectors
- Multiple namespaces
- List vectors

## Type of Change

## Test Plan
Make sure the CI workflow passes for this PR.

If you'd like to run the local integration tests locally, you will need to download Docker, and use `docker compose up` to create the same indexes on your local machine. Once you've done that, you'll need to export environment variables for the different index addresses, along with the dimension.

Here's an example:

```bash
export PINECONE_INDEX_URL_POD="http://localhost:5082" PINECONE_INDEX_URL_SERVERLESS="http://localhost:5081" PINECONE_DIMENSION="1536"
```

Then, you can run the local integration tests directly:
```bash
go test -count=1 -v ./pinecone -run TestRunLocalIntegrationSuite -tags=localServer
```

---
  - https://app.asana.com/0/0/1208056331944375

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,12 +36,13 @@ jobs:
         run: |
           go get ./pinecone
       - name: Run tests
+        continue-on-error: true
         run: go test -count=1 -v ./pinecone
         env:
           PINECONE_API_KEY: ${{ secrets.API_KEY }}
       - name: Run local integration tests
         run: go test -count =1 -v ./pinecone -tags=localServer
         env:
-          PINECONE_INDEX_URL_POD: https://localhost:5081
-          PINECONE_INDEX_URL_SERVERLESS: https://localhost:5082
+          PINECONE_INDEX_URL_POD: http://pc-index-pod:5082
+          PINECONE_INDEX_URL_SERVERLESS: http://pc-index-serverless:5081
           PINECONE_DIMENSION: 1536
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,6 +7,25 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    services:
+      pc-index-serverless:
+        image: ghcr.io/pinecone-io/pinecone-index:latest
+        ports:
+          - 5081:5081
+        env:
+          PORT: 5081
+          DIMENSION: 1536
+          METRIC: cosine
+          INDEX_TYPE: serverless
+      pc-index-pod:
+        image: ghcr.io/pinecone-io/pinecone-index:latest
+        ports:
+          - 5082:5082
+        env:
+          PORT: 5082
+          DIMENSION: 1536
+          METRIC: cosine
+          INDEX_TYPE: pod
     steps:
       - uses: actions/checkout@v4
       - name: Setup Go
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -17,6 +36,13 @@ jobs:
         run: |
           go get ./pinecone
       - name: Run tests
+        continue-on-error: true
```
austin-denoble, line 39: We still want to run the subsequent step that specifically runs local integration tests.

pinecone/local_test.go
```diff
@@ -0,0 +1,154 @@
+//go:build localServer
```
austin-denoble, line 1: Isolating these tests from the other tests. I think this works for not running these tests as a part of the unit / integration test suite, but there may be a better way to do this in Go.

pinecone/test_suite.go
```diff
@@ -54,7 +54,6 @@ func (ts *IntegrationTests) SetupSuite() {
 	for i, v := range vectors {
 		vectorIds[i] = v.Id
 	}
-	ts.vectorIds = append(ts.vectorIds, vectorIds...)
```
austin-denoble, line 57: This was being done twice. I think it's better to keep the actual appending where we upsert successfully. It's inside the `upsertVectors` helper function.

haruska (approved):
:+1: 

This is a good setup. It'll be great if we can eventually eliminate some of the integration testing against prod db.

pinecone/index_connection.go
```diff
@@ -387,9 +387,16 @@ func (idx *IndexConnection) ListVectors(ctx context.Context, in *ListVectorsRequ
...
+	if res.Usage != nil {
+		usage = &Usage{ReadUnits: derefOrDefault(res.Usage.ReadUnits, 0)}
+	} else {
+		usage = nil
+	}
+
 	return &ListVectorsResponse{
 		VectorIds:           vectorIds,
-		Usage:               &Usage{ReadUnits: derefOrDefault(res.Usage.ReadUnits, 0)},
+		Usage:               usage,
```
austin-denoble, line 399: This was actually a nil pointer reference bug discovered from testing against Pinecone Local. Based on the spec and struct, it's possible for `Usage` to be nil, and we weren't properly taking that into account here.

haruska: good catch

PR #79, opened 2024-10-10
Implement Bulk Import, Regenerate core for `2024-10` API

austin-denoble:
## Problem
We are releasing a new version of the API this month: `2024-10`.

There are 3 primary new features that are included in this release:
- Import
- Inference
  - Embed
  - Rerank

This PR implements the operations to support import. Sorry about the size, but you can basically ignore all of the generated code under  `internal/gen` unless you're curious about the new structure of the generated core files. Follow the `codgen/build-clients.sh` script for those details.

## Solution
Since the import operations are technically part of the data plane but only supported via REST, they are represented in the OpenAPI spec and not our protos file. Because of this, we need to change a few things to support these operations `Client` and `IndexConnection` structs to support these operations because traditionally the code `IndexConnection` wraps was targeting gRPC-only db data operations. We now need to generate rest code for the data plane as well so we can interact with imports.

- Update the `codegen/build-clients.sh` script to handle building new modules for both `internal/gen/db_data/grpc` and `internal/gen/db_data/rest`.
- Update `Client` struct and move `NewClientBaseParams` into a field that can be shared more easily when constructing the `IndexConnection`.
  - Add `buildDataClientBaseOptions` to handle constructing the necessary rest client options for the underlying `dbDataClient`.
  - Add an `ensureHostHasHttps` helper as we need to make sure this is present for the index `Host` that's passed, which was not necessary for grpc.
  - Update `Index` method to handle calling `buildDataClientBaseOptions` and passes the new client into `newIndexConnection`.
- Update `IndexConnection` to support both REST and gRPC interfaces under the hood (`restClient`, `grpcClient`).
  - Update `newIndexConnection` to support attaching the new `restClient` to the `IndexConnection` struct.
- Update  `IndexConnection` to support all import operations: `StartImport`, `ListImports`, `DescribeImport`, `CancelImport`.
- Add end-to-end integration test for validating the import flow against serverless indexes.
- Some nitpicky code cleanup, renaming of things around the new rest vs. grpc paradigm, etc.

## Type of Change

## Test Plan
`just test` - make sure CI passes

To see examples of how to use the new methods, check the doc comments.

---
  - https://app.asana.com/0/0/1208325183834377
  - https://app.asana.com/0/0/1208541827330963


haruska (approved):

jhamon (approved):
Great work!

pinecone/client.go
```diff
@@ -32,13 +33,10 @@ import (
...
 // Fields:
 //   - Inference: An InferenceService object that exposes methods for interacting with the Pinecone [Inference API].
-//   - headers: An optional map of HTTP headers to include in each API request, provided through
-//     NewClientParams.Headers or NewClientBaseParams.Headers.
 //   - restClient: Optional underlying *http.Client object used to communicate with the Pinecone API,
 //     provided through NewClientParams.RestClient or NewClientBaseParams.RestClient. If not provided,
 //     a default client is created for you.
-//   - sourceTag: An optional string used to help Pinecone attribute API activity, provided through NewClientParams.SourceTag
-//     or NewClientBaseParams.SourceTag.
+//   - baseParams: A NewClientBaseParams object that holds the configuration for the Pinecone client.
```
aulorbe, line 39: I think you might need to update the example code in this docstring to be `clientParams := pinecone.NewClientBaseParams{}` obj, no?

pinecone/index_connection.go
```diff
@@ -3,12 +3,16 @@ package pinecone
...
+	"encoding/json"
 	"fmt"
+	"io"
 	"log"
+	"net/http"
 	"net/url"
 	"strings"
 
-	"github.com/pinecone-io/go-pinecone/internal/gen/data"
+	db_data_grpc "github.com/pinecone-io/go-pinecone/internal/gen/db_data/grpc"
```
aulorbe, line 14: Question: why are some things snake-cased (`db_data_rest`) while others are camel-cased (`dbDataClient`)?

pinecone/index_connection.go
```diff
@@ -996,8 +1003,307 @@ func (idx *IndexConnection) DescribeIndexStatsFiltered(ctx context.Context, meta
...
+// StartImportResponse holds the response parameters for the StartImport method.
+//
+// Fields:
+//   - Id: The ID of the import process that was started.
+type StartImportResponse struct {
+	Id string `json:"id,omitempty"`
+}
+
+// StartImport imports data from a storage provider into an index. The uri parameter must start with the
+// schema of a supported storage provider. For buckets that are not publicly readable, you will also need to
```
aulorbe, line 1015: Nit: I'd add `e.g. "s3://"` as an example of the start of the uri param, just b/c some ppl might not know what that means

pinecone/index_connection.go
```diff
@@ -996,8 +1003,307 @@ func (idx *IndexConnection) DescribeIndexStatsFiltered(ctx context.Context, meta
...
+// separately configure a storage integration and pass the integration id.
```
aulorbe, line 1016: Do we have instructions in our docs for how to build a storage integration? If we're telling ppl they need to do an extra step, it might be nice for us to provide them resources about how to do that in the same sentence

pinecone/index_connection.go
```diff
@@ -996,8 +1003,307 @@ func (idx *IndexConnection) DescribeIndexStatsFiltered(ctx context.Context, meta
...
+//
+// Returns a pointer to a StartImportResponse object with the import ID or an error if the request fails.
+//
+// Parameters:
+//   - ctx: A context.Context object controls the request's lifetime,
+//     allowing for the request to be canceled or to timeout according to the context's deadline.
+//   - uri: The URI of the data to import. The URI must start with the scheme of a supported storage provider.
```
aulorbe, line 1023: `scheme` >> `schema`

pinecone/index_connection.go
```diff
@@ -996,8 +1003,307 @@ func (idx *IndexConnection) DescribeIndexStatsFiltered(ctx context.Context, meta
...
+
+// DescribeImport retrieves information about a specific import operation.
+//
+// Returns an Import object representing the current state of the import or an error if the request fails.
+//
+// Parameters:
+//   - ctx: A context.Context object controls the request's lifetime,
+//     allowing for the request to be canceled or to timeout according to the context's deadline.
+//   - id: The id of the import operation. This is returned when you call StartImport, or can be retrieved
+//     through the ListImports method.
```
aulorbe, line 1097: Can you link to other funcs in the docstrings of Go funcs? Might be nice to link to `ListImports` here

pinecone/index_connection_test.go
```diff
@@ -269,8 +269,34 @@ func (ts *IntegrationTests) TestUpdateVectorSparseValues() error {
 	actualSparseValues := vector.Vectors[ts.vectorIds[0]].SparseValues.Values
 
 	assert.ElementsMatch(ts.T(), expectedSparseValues.Values, actualSparseValues, "Sparse values do not match")
+}
+
+func (ts *IntegrationTests) TestImportFlow() {
```
aulorbe, line 274: Might want to do some negative examples (like confirming errors are thrown) too

pinecone/test_suite.go
```diff
@@ -87,7 +87,9 @@ func (ts *IntegrationTests) TearDownSuite() {
 	_, err = WaitUntilIndexReady(ts, ctx)
 	require.NoError(ts.T(), err)
 	err = ts.client.DeleteIndex(ctx, ts.idxName)
-	require.NoError(ts.T(), err)
+	if err != nil {
```
aulorbe, line 90: Do you need to retry here? I know sometimes w/configure index calls and the like, we have to retry in the TS client

pinecone/client.go
austin-denoble: You don't need to provide `pinecone.NewClientBaseParams{}`, you should be able to instantiate and interact with the client in the same way as before using `pinecone.NewClientParams{}`. I've just added a private field `baseParams` to bundle that up and reuse it when creating an index connection. There should be no visible impact to the user.

pinecone/index_connection.go
austin-denoble: Added a link to: https://docs.pinecone.io/guides/operations/integrations/manage-storage-integrations

pinecone/index_connection.go
```diff
@@ -996,8 +1003,307 @@ func (idx *IndexConnection) DescribeIndexStatsFiltered(ctx context.Context, meta
...
+// separately configure a storage integration and pass the integration id.
+//
+// Returns a pointer to a StartImportResponse object with the import ID or an error if the request fails.
+//
+// Parameters:
+//   - ctx: A context.Context object controls the request's lifetime,
+//     allowing for the request to be canceled or to timeout according to the context's deadline.
+//   - uri: The URI of the data to import. The URI must start with the scheme of a supported storage provider.
+//   - integrationId: If your bucket requires authentication to access, you need to pass the id of your storage integration using this property.
+//     Pass nil if not required.
```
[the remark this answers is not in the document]
austin-denoble: I would need to add a new struct type that allows batching `uri`, `integrationId`, and `errorMode`. I could change it but I felt like since there are so few fields at the moment maybe it's more overhead to have to worry about creating the struct rather than just passing values. Do you feel strongly about this?

pinecone/index_connection.go
```diff
@@ -996,8 +1003,307 @@ func (idx *IndexConnection) DescribeIndexStatsFiltered(ctx context.Context, meta
...
+//   - errorMode: If set to "continue", the import operation will continue even if some records fail to import.
```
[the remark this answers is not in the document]
austin-denoble: Yes, the full doc for `errorMode` looks like this:

```
errorMode: If set to "continue", the import operation will continue even if some records fail to import.
Pass "abort" to stop the import operation if any records fail. Will default to "continue" if nil is passed.
```

pinecone/index_connection.go
austin-denoble: Good call, updated to link to the explicit methods. I think we could do more of this elsewhere, I think we've just been using proper names in the doc comments until now. Added a chore ticket to go through these: https://app.asana.com/0/1203260648987893/1208561704179017/f

pinecone/index_connection.go
austin-denoble: I think it's supposed to be scheme, right? https://en.wikipedia.org/wiki/List_of_URI_schemes

pinecone/client.go
aulorbe: Ahh okay thank you!

pinecone/index_connection.go
aulorbe: ah cool !

pinecone/index_connection.go
aulorbe: oh idk, you just say `schema` above, so I thought it should be `schema` :)

pinecone/index_connection.go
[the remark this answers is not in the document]
aulorbe: no def not, you do you!

pinecone/index_connection.go
aulorbe: Sweet sounds great, thanks!

pinecone/test_suite.go
austin-denoble: Added a small retry mechanism here for cleanup.

pinecone/index_connection_test.go
austin-denoble: I confirmed that we're returning an error when trying to call `StartImport` with an empty URI.

PR #86, opened 2024-10-23
Regenerate client code for `2024-10`

austin-denoble:
## Problem
PR was merged today making some slight adjustments to the db control spec to better align with the server impl: https://github.com/pinecone-io/apis/pull/181

## Solution
Slightly adjust the `PodSpec` type and related code to account for the new shape.

## Type of Change

## Test Plan
`just test` - CI

### DIFF
--- a/pinecone/client.go
+++ b/pinecone/client.go
@@ -1509,9 +1509,9 @@ func toIndex(idx *db_control.IndexModel) *Index {
 		spec.Pod = &PodSpec{
 			Environment:      idx.Spec.Pod.Environment,
 			PodType:          idx.Spec.Pod.PodType,
-			PodCount:         derefOrDefault(idx.Spec.Pod.Pods, 0),
-			Replicas:         derefOrDefault(idx.Spec.Pod.Replicas, 0),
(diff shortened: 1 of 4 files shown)


pinecone/client.go
```diff
@@ -1506,9 +1509,9 @@ func toIndex(idx *db_control.IndexModel) *Index {
 		spec.Pod = &PodSpec{
 			Environment:      idx.Spec.Pod.Environment,
 			PodType:          idx.Spec.Pod.PodType,
-			PodCount:         int32(idx.Spec.Pod.Pods),
-			Replicas:         idx.Spec.Pod.Replicas,
-			ShardCount:       idx.Spec.Pod.Shards,
+			PodCount:         derefOrDefault(idx.Spec.Pod.Pods, 0),
```
austin-denoble, line 1512: I don't think these should ever realistically be `nil` when coming back from the server.

haruska: I agree... maybe the default should be 1 (as that's the count if all other defaults are used)

haruska (approved):

PR #89, opened 2024-12-20
Update `ConfigureIndex` to merge new and previous `IndexTags`, update `README` examples

austin-denoble:
## Problem
@jhamon called out that since `ConfigureIndex` is a PATCH operation, we should probably be explicitly merging existing tags with incoming tags when configuring an index as a partial update. See here for a python example: https://github.com/pinecone-io/pinecone-python-client/pull/426/files#diff-9e7f01373e4589c7e4ded21935023cdf187ab6ce8b2b44a0d5470a27eaf464eeR303-R314

## Solution
- Update `ConfigureIndex` to call `DescribeIndex` and merge tags properly on updates.
- Update `README` to add examples of using `IndexTags` when creating and updating indexes. 

## Type of Change
- [X] Non-code change (docs, etc)

## Test Plan
`just test`
Make sure CI is green.

### DIFF
--- a/pinecone/client.go
+++ b/pinecone/client.go
@@ -934,6 +934,13 @@ func (c *Client) ConfigureIndex(ctx context.Context, name string, in ConfigureIn
 	replicas := pointerOrNil(in.Replicas)
 	deletionProtection := pointerOrNil(in.DeletionProtection)
 
+	// Describe index in order to merge existing tags with incoming tags
+	idxDesc, err := c.DescribeIndex(ctx, name)
+	if err != nil {
+		return nil, err
+	}
+	existingTags := idxDesc.Tags
+
 	var request db_control.ConfigureIndexRequest
 	if podType != nil || replicas != nil {
 		request.Spec =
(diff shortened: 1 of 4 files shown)


pinecone/index_connection_test.go
```diff
@@ -90,7 +90,6 @@ func (ts *IntegrationTests) TestDeleteVectorsByFilter() {
 
 	if ts.indexType == "serverless" {
 		assert.Error(ts.T(), err)
-		assert.Containsf(ts.T(), err.Error(), "Serverless and Starter indexes do not support deleting with metadata filtering", "Expected error message to contain 'Serverless and Starter indexes do not support deleting with metadata filtering'")
```
austin-denoble, line 93: This error message has changed slightly since this test was originally written. I think it's fine to remove the string check for now and just validate that we've gotten an error back.

pinecone/client.go
```diff
@@ -953,8 +960,7 @@ func (c *Client) ConfigureIndex(ctx context.Context, name string, in ConfigureIn
 			}
 	}
 	request.DeletionProtection = (*db_control.DeletionProtection)(deletionProtection)
-	request.Tags = (*db_control.IndexTags)(&in.Tags)
-	fmt.Printf("request.Tags: %+v\n", request.Tags)
```
austin-denoble, line 957: whoops - left this in previously

haruska (approved):

PR #91, opened 2025-01-06
Fix `v2.X.X` releases for the client, update `go.mod`

austin-denoble:
## Problem
We made a mistake when releasing the previous `v2.X.X` of the Go client. In order to handle major versions properly, the Go module ecosystem requires that packages include a `/vX` path for major versions. This needs to be represented in the `go.mod` file for the module. See https://go.dev/doc/modules/version-numbers under "Major version".

Without making this change to the `go.mod` file, when attempting to install >v2.x.x, you end up with an error:

```
go get -u github.com/pinecone-io/go-pinecone/pinecone@v2.0.0
go: github.com/pinecone-io/go-pinecone/pinecone@v2.0.0: github.com/pinecone-io/go-pinecone@v2.0.0: invalid version: module contains a go.mod file, so module path must match major version ("github.com/pinecone-io/go-pinecone/v2")
```

## Solution
- Update `go.mod` for v2 -> `module github.com/pinecone-io/go-pinecone/v2`.
- Update places where we're importing from `github.com/pinecone-io/go-pinecone/internal` to `github.com/pinecone-io/go-pinecone/v2/internal`.

## Type of Change

## Test Plan
We need to tag and release a new `v2.2.0` to test that this works as expected. Testing against previous 2.x.x releases is hampered by the fact that the `go.mod` file is incorrect.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,6 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
 	github.com/oapi-codegen/runtime v1.1.1
-	github.com/pinecone-io/go-pinecone v1.1.1
 	github.com/stretchr/testify v1.8.4
 	google.golang.org/genproto/googleapis/api v0.0.0-20240528184218-531527333157
 	google.golang.org/grpc v1.65.0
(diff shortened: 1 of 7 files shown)


haruska (approved):
build is failing but code change looks fine.

austin-denoble:
> build is failing but code change looks fine.

Seems like some of the ConfigureIndex tests were taking a bit longer than expected. They were a bit flakey before but has gotten worse, so I bumped the sleep time for now.

PR #118, opened 2025-07-17
Add PercentComplete and RecordsImported fields to Import object construction

vadimpanin:
## Problem

`IndexConnection.DescribeImport` returns an `Import` object with `PercentComplete` and `RecordsImported` fields being always 0, because they are not passed to `Import` object construction in `toImport` function.

## Solution

Add `PercentComplete` and `PercentComplete` fields passing to `Import` object construction in `toImport` function.

## Type of Change

## Test Plan

- Start a batch import using `IndexConnection.StartImport`
- Continuously call `IndexConnection.DescribeImport`
- Verify that the returned `Import` object's fields have proper values

### DIFF
--- a/pinecone/index_connection.go
+++ b/pinecone/index_connection.go
@@ -1901,6 +1901,8 @@ func toImport(importModel *db_data_rest.ImportModel) *Import {
 		CreatedAt:  importModel.CreatedAt,
 		FinishedAt: importModel.FinishedAt,
 		Error:      importModel.Error,
+		PercentComplete: derefOrDefault(importModel.PercentComplete, 0),
+		RecordsImported: derefOrDefault(importModel.RecordsImported, 0),
 	}
 }
 


pinecone/index_connection.go
```diff
@@ -1901,6 +1901,8 @@ func toImport(importModel *db_data_rest.ImportModel) *Import {
 		CreatedAt:  importModel.CreatedAt,
 		FinishedAt: importModel.FinishedAt,
 		Error:      importModel.Error,
+		PercentComplete: *importModel.PercentComplete,
+		RecordsImported: *importModel.RecordsImported,
```
austin-denoble, line 1905: Rather than directly dereferencing, could you add the `derefOrDefault` function to each of these, with a default of 0? There are examples in the test PR I posted: https://github.com/pinecone-io/go-pinecone/pull/119/files#diff-8fa19c4bc5c00a4e22fb57af19d984e15324c10666951ba321ef441ff12a480fR1904-R1905

It's possible there are instances where the server may return nil for these pointers, so we need to be a little defensive.

austin-denoble (changes requested):

